PR #31, opened 2021-02-24
Fix charts tooltip

### LINKED ISSUE #7: [charts] Tooltip is not at mouse position
When hovering over the points of a line chart, the tooltip is sometimes at the center line, sometimes somewhere else - would probably be best to always keep it at mouse position?

komninoschat:
Fixes #7.


silvester-pari (approved):
Looks great, thanks! Good work!

PR #45, opened 2021-03-24
add programmatic hover

### LINKED ISSUE #38: [charts] allow programmatic hover on chart
When e.g. the current selected date is changed in the app somewhere, the chart should be able to react and also show the current selected date as highlighted

komninoschat:
Fixes #38

### DIFF
--- a/docs/.vuepress/components/ChartsAdvanced.vue
+++ b/docs/.vuepress/components/ChartsAdvanced.vue
@@ -1,16 +1,28 @@
 <template>
-  <charts
-    type="line"
-    :dataObject='dataObject'
-    :plotConfig='plotConfig'
-    :indexTooltips='true'
-    style="width: 100%; height: 100%;"
-    @dateSelected='dateSelected'
-  />
+  <div>
+    <charts
+      type="line"
+      :dataObject='dataObject'
+      :plotConfig='plotConfig'
+      :indexTooltips='true'
+      style="width: 100%; height: 100%;"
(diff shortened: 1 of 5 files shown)


silvester-pari:
Thanks @komninoschat! Unfortunately I didn't manage to get it working :thinking: 
Could you update the `ChartsAdvanced.vue` example so I can see how it is intended to work? E.g. with a date in December 2018.

There is also a linter error preventing build btw.

komninoschat:
Regarding the status of this PR:

1. I managed to programmatically highlight the dots on the graphs
2. In our current `charts.js` version(2.9.3), there is no API to programmatically show tooltips(and lines on the advanced graph which depend on them)

We have two options for (2):

1. Upgrade to `charts.js` 3.0.0. [This](https://github.com/chartjs/Chart.js/pull/7845) PR which was merged on version 3.0.0-beta.5 adds the API we want. Also, this major release has been in the works for quite some time now and they are already on release candidate 6 so it's relatively stable. The downside to this is that we will have to immediately deal with the breaking changes.
2. Use the hack solution @silvester-pari wrote above.

I **strongly** suggest we follow (1) in order to reduce our technical debt, instead of further increasing it.

silvester-pari:
Great find, thanks! I think we should do 1) as you proposed!

silvester-pari:
Unfortunately, chartjs 3 is not yet compatible with the Vue wrapper (I think). Maybe it's best to use our original hack as described in #38 until it becomes available.

Issue #34, opened 2021-02-25
[map-tooltip] Adapt tooltip behaviour in map to match charts


silvester-pari:
Inside of charts, the tooltip now stays inside of boundaries and switches side when going too far right (thanks to https://github.com/EOX-A/elements/pull/31). A similiar behaviour should be implemented when used on the map aswell.

PR #49, opened 2021-03-26
fix tooltip behaviour

### LINKED ISSUE #33: [map-tooltip] Tooltip re-renders multiple times inside of feature
With the new openlayers version (see also #32), the tooltip behaviour has changed: the hover is somehow now firing mulltiple times, also inside of the same vector feature. Each time the tooltip is rendered and destroyed again, resulting in an unpleasant re-opening when moving the mouse.

komninoschat:
Fixes #33, #34


silvester-pari (approved):
thanks for this, works well!

PR #59, opened 2021-05-03
Automatically stretch chart height

### LINKED ISSUE #57: [charts] make charts responsive (height/width)
right now, only the width adapts to the parent container, would be nice if the height does the same.

maybe set some aspect ratio limits if the ratio gets too extreme (optional)

radupasparuga:
Fixes #57


silvester-pari:
Great, thanks!

silvester-pari (approved):

PR #65, opened 2021-09-16
feat: Add locale to eox/chart

### LINKED ISSUE #63: [charts] Add chart internationalization
Especially for the xAxis showing times, there should be an option to pass a language or a way to replace strings.

Outcome should be that the times in the xAxis should be displayable in english, german, etc.

radupasparuga:
Fixes #63


silvester-pari (approved):
Great, thanks! LGTM

PR #84, opened 2023-05-11
[chart] Chart Events

silvester-pari:
feat: signalsmanager emits events when failing to fetch data or when retrying to fetch data


silvester-pari:
Has been solved otherwise, plus conflicts with ne web-component setup. CLosing for now!

Issue #138, opened 2023-07-02
Add default map controls


silvester-pari:
For those cases where the hosting app does not provide any UI. Basic map controls would be zoom & attribution. Should be configurable via attribution, e.g. `<eox-map controls='["zoom", "attribution"]'></eox-map>`.

Issue #131, opened 2023-07-02
Use lat/lon coordinate format as default for attributes


silvester-pari:
The center attribute could accept lat/lon and mercator and could try to guess which one is used. Maybe there is something usable from OL already

PR #147, opened 2023-07-05
parse webcomponent attributes

RobertOrthofer:
this PR implements the the parsing of lat/lon-coordinates as defined in https://github.com/EOX-A/EOxElements/issues/131 and the default controls as defined in https://github.com/EOX-A/EOxElements/issues/138

The "guessing" part of the center coordinates is adapted from `ol`, where already a warning is emitted when trying to set the center to a coordinate that "most likely" is lont/lat. This behaviour can always be overruled by manually setting the center of the view by calling `view.setCenter()`

Controls are added by setting an array of ol-control-names. The id for later removal is equal to the name of the control. Most likely only one instance of any control should be present on a map at time.


silvester-pari (approved):
Great, thanks for your work! Looks really good and works well. I am just wondering regarding the controls, if/how we should allow passing options to the control creation. Things like defining the OverviewMap layer(s), class names, tooltip labels etc. But this could be seen as an evolution for another PR, as the initial requirement was only zoom and attribution.

Issue #111, opened 2023-06-21
setTileUrlFunction() adds up and is called multiple times


silvester-pari:
We are [setting the tile URL function with setTileUrlFunction](https://github.com/EOX-A/EOxElements/blob/225a0a17da869ec583ddb6da2d97a150d650ec31/elements/timecontrol/src/main.ts#L128), but this somehow is stacked up and when logging inside the function the first time it is called 2 times, then 4 times, then 6 times, etc.

It still works, but will probably cause some memory problems if not handled.

silvester-pari:
Maybe @RobertOrthofer you could have a look at this? Might need some other way to do this, as it is a hack currently...

PR #152, opened 2023-07-21
fix: set original tileUrlFunction only once

RobertOrthofer:
This PR makes sure that the original TileUrlFunction (responsible for getting the correct tile code) is not altered along the way, thus creating memory leaks. 

The `render` function is called every time step, i am uncertain if this is the correct behaviour, or if this is due to the testing setup.

Fixes https://github.com/EOX-A/EOxElements/issues/111

### DIFF
--- a/elements/timecontrol/src/main.ts
+++ b/elements/timecontrol/src/main.ts
@@ -55,7 +55,6 @@ export class EOxDrawTools extends LitElement {
   @property()
   urlFunction: Function;
 
-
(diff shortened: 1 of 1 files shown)


silvester-pari (approved):
Great stuff! Thanks for this.

The `render()` function is called every time the component is updated, and it re-renders the changed bits ( in our case e.g. the displayed date, the progress of the slider etc.).

Issue #139, opened 2023-07-02
Add minimap configuration


silvester-pari:
A minimap should be configurable via attributes, e.g. `<eox-map minimap>` (maybe using OSM as layer, or the first layer of the map) or `<eox-map minimap="layer_id">`.

Alternatively, the minimap could be a dedicated web component that is hosted inside a slot/template to give more styling flexibility (as with tooltips).

silvester-pari:
https://github.com/EOX-A/EOxElements/pull/147 introduces a clean way to add the minimap control, but what is missing is a way to set the config for it (especially what layer(s) should be displayed on the minimap).

Either the `controls` attribute could be expanded to allow objects, or we introduce a new attribute for e.g. configuring the minimap, and it is switched on/off via the `controls` attribute. Not sure yet what would be best.

PR #153, opened 2023-07-24
Controls as JSON

RobertOrthofer:
This PR implements the controls as JSON. The Key of each property is the name of the control, the property itself represents the `options` of the ol-control.

If the control has a `layers`-Option (e.G. the `OverviewMap`), a JSON is expected, in the same way as when defining layers. Those layers will get parsed and will be added to the control only. The test is updated to display an `OverviewMap`-Control with OSM Tiles, which are not represented in the main map.

### DIFF
--- a/elements/map/src/controls.ts
+++ b/elements/map/src/controls.ts
@@ -1,5 +1,6 @@
 import { EOxMap } from "../main";
 import * as olControls from "ol/control";
+import { generateLayers } from "./generate";
 
 type controlType =
   | "Attribution"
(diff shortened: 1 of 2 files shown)


silvester-pari (approved):
Nice! Thank you. For those controls that don't need any options (e.g. the user wants to add zoom, attribution and fullscreen with no options) it would be nice if the "old" array format would still be accepted - but maybe this is too much effort from a TypeScript perspective (as the parsed JSON could be either an `Array<controlType>` or a `controlDictionary`).
I think for the first iteration this is already a great solution!

Closes https://github.com/EOX-A/EOxElements/issues/139

PR #155, opened 2023-07-24
Controls as Array

RobertOrthofer:
This PR restores the possibility to define simple controls (without options) as Array.


silvester-pari (approved):
That was quick. Awesome, thanks!

PR #158, opened 2023-07-24
Layercontrol/feature/groups

silvester-pari:
This PR introduces the functionality to render and modify layer groups, whereas each group can be visibility toggled/opacity changed/sorted like a layer.
It is currently only possible to sort groups (1 level deep) and layers, but not drag/drop one layer of one group into another etc.

WIP / TODO: 
- [ ] zIndex sorting/display has been disabled for this to work --> reenable
- [x] optional layers are added on top of list instead of on top of respective group -> should be inserted at least in correct group


silvester-pari:
Regarding zIndex, I think there is no easy way to handle it in conjunction with groups. What we could do is e.g. only respect the zIndex on the top-level, but that would kinda defeat the purpose of zIndex. Commenting it out for now until we find a solution for this.

PR #167, opened 2023-07-27
Spatial filtering

### LINKED ISSUE #141: Add spatial filtering
Allow the user to draw a polygon/bbox (using `eox-map`) to filter items.
This could use some turf package for frontend filtering, and send the bbox to a backend in external mode.

silvester-pari:
This PR adds spatial filtering to the itemfilter by using an `eox-map` in a slot. It uses some packages from `turf` to perform the filtering, but this might be replaced by some OpenLayers functionalities (since we load OpenLayers anyways with `eox-map`).

Closes #141


silvester-pari:
This functionality uses three packages from turf.js [here](https://github.com/EOX-A/EOxElements/pull/167/files#diff-27d790f9f7ffb07a51125dc99f4b5060482691ea8e4f93f09e4108b48d1cd539R10) and [here](https://github.com/EOX-A/EOxElements/pull/167/files#diff-27d790f9f7ffb07a51125dc99f4b5060482691ea8e4f93f09e4108b48d1cd539R14) in order to 
1) convert a bbox into a polygon (for convenience, could be done manually), and then to
2) check if two polygons a) intersect or b) one contains the other

I have seen that OL offers similar functionality, with "filters", e.g. [intersects](https://openlayers.org/en/latest/apidoc/module-ol_format_filter.html#.intersects) and [within](https://openlayers.org/en/latest/apidoc/module-ol_format_filter.html#.within). 

Still, I have opted against using the OL functionality (i.e. via a helper function provided by the eox-map), since there might be cases where we offer spatial sorting/filtering inside the itemfilter without a map. Example: hardcoded or dynamic (GPS sensor) user position, and filter items within a radius / sort items by distance, without offering a map interface.

I would like your opinion @RobertOrthofer @santilland if that also makes sense for you, if so then I will merge this within the next days.

RobertOrthofer:
Using `turf` for geometric operations is definitely the go-to method, even with ol being loaded. There used to be issues with bloating up the bundle size, but these issues are resolved now as far as I am aware.

some tests are still failing for me, trying to find an `[data-cy='expand-button']`, I suppose this is not part of this PR?

RobertOrthofer (approved):

santilland:
I think it sounds sensible to use turf as well, so also from me a go ahead. Probably best to have a second look at the tests as Robert was saying.

silvester-pari:
Thanks for your comments! I simplified it a bit (and at the same time made it more flexible) and removed one of the turf packages by using geometries instead of just bboxes.

PR #171, opened 2023-07-28
Style Improvements II

spectrachrome:


santilland:
From the checks it seems prettier needs to be run on the files, not sure how the automatic bump version is configured, but it seems it is not correctly configured for the new eox_styles element

spectrachrome:
Code is now formatted with Prettier, since there is no `eox_styles` element (just CSS and a few modifications in pre-existig components), there is nothing new to version-bump - so this must be a pre-existing condition here.

silvester-pari (approved):
Thank you, looks good!

PR #186, opened 2023-08-08
Implement time-aware slider ticks as found in `eodash`

spectrachrome:
This pull request takes the slider ticks implementation from `eodash` and translates it into a Lit component so that it can also easily be used by other applications.

https://github.com/EOX-A/EOxElements/assets/94269527/3180ac4c-ca88-4f1b-990e-859ef75653eb

The logic generating the year marks has been fixed and is now working properly.


silvester-pari (approved):
Great addition, thanks!

PR #208, opened 2023-08-22
Add ol-stac

silvester-pari:
This PR adds the ol-stac package for automatic rendering of STAC layers. Thanks @m-mohr for this package!

### DIFF
--- a/elements/map/package.json
+++ b/elements/map/package.json
@@ -6,7 +6,7 @@
     "lit": "^2.7.4",
     "ol": "^7.4.0",
     "ol-mapbox-style": "^10.6.0",
-    "ol-stac": "^1.0.0-beta.1"
+    "ol-stac": "^1.0.0-beta.3"
   },
   "devDependencies": {
     "@eox/eslint-config": "^1.0.0",
--- a/elements/map/package.json
+++ b/elements/map/package.json
@@ -6,7 +6,7 @@
     "lit": "^2.7.4",
     "ol": "^7.4.0",
     "ol-mapbox-style": "^10.6.0",
-    "ol-stac": "^1.0.0-beta.3"
+    "ol-stac": "^1.0.0-beta.4"
   },
   "devDependencies": {
     "@eox/eslint-config": "^1.0.0",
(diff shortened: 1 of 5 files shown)


RobertOrthofer:
I added a simple test for the STAC-Layer, using an URL from an [example](https://m-mohr.github.io/ol-stac/en/latest/examples/stac-item.html).
It looks like the properties are not passed to the super constructor of the `LayerGroup`, which the `StacLayer` ist extending. This functionality is needed here because we rely on the `id` to be set via our JSON structure.

elements/map/src/generate.ts
```diff
@@ -1,10 +1,20 @@
 import * as olLayers from "ol/layer";
 import * as olSources from "ol/source";
 import * as olFormats from "ol/format";
+import STAC from "ol-stac/layer/STAC";
```
m-mohr, line 4:
```suggestion
import STAC from "ol-stac";
```

elements/map/package.json
```diff
@@ -5,7 +5,8 @@
   "dependencies": {
     "lit": "^2.7.4",
     "ol": "^7.4.0",
-    "ol-mapbox-style": "^10.6.0"
+    "ol-mapbox-style": "^10.6.0",
+    "ol-stac": "^1.0.0-beta.1"
```
m-mohr, line 9:
```suggestion
    "ol-stac": "^1.0.0-beta.3"
```

elements/map/src/generate.ts
RobertOrthofer: This does not seem to work, while the properties work perfectly fine now via `import STAC from "ol-stac/layer/STAC";`

m-mohr: Hmm, thanks. I'll check it.

RobertOrthofer (approved):
thanks @m-mohr for your work on the library, looks good to me!

elements/map/src/generate.ts
silvester-pari: I think it's this line that still creates a link to `index.js` https://github.com/m-mohr/ol-stac/blob/main/tasks/prepare-package.js#L13

m-mohr:
Yeah, thanks, just found that spot, too. I released beta.4. Does it work for you now?

elements/map/src/generate.ts
silvester-pari: `beta4` fixed it! :tada:

Issue #151, opened 2023-07-19
getLayerById inside layer groups


silvester-pari:
https://github.com/EOX-A/EOxElements/commit/00e98f17bdc86e692548745e783ce8efdd69dd5e introduced layer group generation via JSON. The map helper function `getLayerById` does not yet find layers that are inside groups; this can be tested with e.g. the following layers:

```
[
  {
    type: "Group",
    layers: [
      {
        type: "Vector",
        id: "regions",
        source: {
          type: "Vector",
          url: "https://openlayers.org/data/vector/ecoregions.json",
          format: "GeoJSON",
        },
      },
    ],
  },
  {
    type: "Tile",
    source: {
      type: "OSM",
    },
  },
]
```

Output: 
```
Uncaught Error: Layer with id: regions does not exist.
```

silvester-pari:
Some brainstorming on how to get a layer by id inside one or multiple nested groups:
```
const findLayer = (layers, id) => {
  let found;
  const search = (searchLayers, searchId) => {
    found = searchLayers.find((layer) => layer.get("id") === searchId);
    if (found) {
      return found;
    } else {
      const groups = searchLayers.filter((layer) => !!layer.getLayers);
      groups.forEach((group) => search(group.getLayers().getArray(), id));
    }
  };
  search(layers, id);
  return found;
};

findLayer(map.getLayers().getArray(), "myLayer");
```

silvester-pari:
After some more experimentation, I came up with the following setup, shich not only allows getting one layer by id, but also filtering for other properties:

```
  filterLayers = (
    layers: Array<Layer | LayerGroup>,
    key: string,
    value: any
  ) => {
    let found: Array<Layer | LayerGroup> = [];
    const search = (
      searchLayers: Array<Layer | LayerGroup>,
      key: string,
      value: any
    ) => {
      found = [...found, ...searchLayers.filter((l) => l.get(key) === value)];
      // @ts-ignore
      const groups = searchLayers.filter((l) => l.getLayers);
      if (groups.length > 0) {
        groups.forEach((group) =>
          // @ts-ignore
          search(group.getLayers().getArray(), key, value)
        );
      }
      return found;
    };
    search(layers, key, value);
    return found;
  };

  findLayerById = (layers: Array<Layer | LayerGroup>, id: string) => {
    return this.filterLayers(layers, "id", id)[0];
  };
```

[...]

PR #217, opened 2023-08-24
get layer inside group

RobertOrthofer:
This PR is based in the ideas listed in #151. In order to get all the layers inside groups, but also the group layers itself, a function `getFlatLayersArray` was created. Maybe we want to expose this helper function, as filtering for attributes, but also non-property-based attributes (like `visible` to get all visible layers) would be 1-liners in the parent application?

The `getLayerById` function remained mostly unchanged, as it contains some special logic for mapbox-layers, which may or may not be used in the future. It currently also throws an error, which was useful when testing, but maybe it should just return `undefined` instead?

@silvester-pari currently the `generate`- function does a `Array.reverse` on the layers array, is there a certain reason for that? the added test does not show the vector layer, although it should be showing considering the layer order.

### DIFF
--- a/elements/map/src/layer.ts
+++ b/elements/map/src/layer.ts
@@ -1,4 +1,6 @@
 import { EOxMap } from "../main";
+import Group from "ol/layer/Group";
+import Layer from "ol/layer/Base";
 
 /**
  *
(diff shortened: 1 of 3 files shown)


silvester-pari:
Thanks @RobertOrthofer , the `reverse()` is because this way, the ordering of the layers in the layer JSON is equal to the display on the map (thinking of "higher up" in the JSON equals to "higher up" in the map). But maybe this needs to be reconsidered. Looking at your changes in the meantime.

elements/map/src/layer.ts
```diff
@@ -7,16 +9,38 @@ import { EOxMap } from "../main";
...
 export function getLayerById(EOxMap: EOxMap, layerId: string) {
-  // get mapbox-style layer or manually generated ol layer
-  const layers = EOxMap.map.getLayers().getArray();
+  const flatLayers = getFlatLayersArray(EOxMap.map.getLayers().getArray());
+  // get mapbox-style layer or manually generated ol layer, both group or regular layers
+
   const layer =
-    layers.find((l) => l.get("id") === layerId) ||
-    layers
+    flatLayers.find((l) => l.get("id") === layerId) ||
```
silvester-pari, line 16: Maybe this could be overrideable (e.g. as third, optional, function parameter), because there could be some edge cases where it's not "id"?

silvester-pari (changes requested):
+1 for exposing the `flatLayersArray` function, sounds like a good idea.

PR #220, opened 2023-08-28
chore: improve typing

RobertOrthofer:
This PR upgrades to ol 8 (which brings native style expressions) and improves the use of TypeScript in the Map Element.

The removal of the unneeded generic `Function`-Typing in `main.ts` allows us to receive the correct types in the parent application.

`layer` and `source` types are now done manually, defining all relevant types of layers and sources as strings. Before TS expected actual ol objects, which was not the intention.

The options-objects from `ol` can be utilised by importing the type (e.g. `import('ol/interaction/Draw').Options`), and adapting them for our needs, either by adding properties via the `&` operator and overloading the object with new properties, or omitting some properties via `Omit`, or doing both to change the type of a property.

### DIFF
--- a/elements/map/src/generate.ts
+++ b/elements/map/src/generate.ts
@@ -5,6 +5,7 @@ import STAC from "ol-stac";
 import { applyStyle } from "ol-mapbox-style";
 import { FlatStyleLike } from "ol/style/flat";
 import mapboxgl, { AnySourceData } from "mapbox-gl";
+import { Collection } from "ol";
 
 const availableLayers = {
   ...olLayers,
@@ -58,9 +59,8 @@ export type EoxLayer = {
   style?: mapboxgl.Style | FlatStyleLike;
 };
 
-export function createLayer(layer: EoxLayer, group?: string): olLayers.Layer {
+export function createLayer(layer: EoxLayer): olLayers.Layer {
   const newLayer = availableLayers[layer.type];
-  //@ts-ignore
   const newSource = availableSources[layer.source?.type];
   if (!newLayer) {
     throw new Error(`Layer type ${layer.type} not supported!`);
@@ -72,7 +72,6 @@ export function createLayer(layer: EoxLayer, group?: string): olLayers.Layer {
   //@ts-ignore
   const olLayer = new newLayer({
     ...layer,
-    group,
     ...(layer.source && {
       //@ts-ignore
       source: new newSource({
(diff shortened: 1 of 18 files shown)


elements/map/src/generate.ts
```diff
@@ -50,9 +83,8 @@ export function createLayer(layer: EoxLayer, group?: string): olLayers.Layer {
       }),
     }),
     style: undefined, // override layer style, apply style after
-    // @ts-ignore
     ...(layer.type === "Group" && {
-      layers: layer.layers.reverse().map((l) => createLayer(l, layer.id)),
+      layers: layer.layers.map((l) => createLayer(l, layer.id)),
```
silvester-pari, line 87: Removing the `reverse()` here will reverse all the current map collections, right? Is there a reason for doing that?

elements/map/src/generate.ts
```diff
@@ -11,23 +11,55 @@ const availableLayers = {
...
   ...olSources,
 };
 
 export type EoxLayer = {
-  type: olLayers.Layer;
-  id: string;
-  properties?: Object;
-  source?: { type: olSources.Source };
+  type: layerType;
+  id?: string;
```
silvester-pari, line 53: I think we can agree to always put the `id` into the `properties` object, to be OL API docs conformant. What do you say?

elements/map/src/generate.ts
RobertOrthofer: this was a mistake, trying to fix the `groupLayer`-Test. I was not aware that the EOX-Map expects the layer order in the opposite way of ol, i will re-add the `reverse` and update the groupLayer-Test

elements/map/src/generate.ts
RobertOrthofer: true, i updated the EoxLayer type

silvester-pari: While we're at it, the `group` id should also move into the `properties` to be OL conform, currently the `group`is added at root level ike the `id` @RobertOrthofer

RobertOrthofer: [60b65c4](https://github.com/EOX-A/EOxElements/pull/220/commits/60b65c4a1172a48daf6720f0e94a368b7fa69380) now sets a reference to the parent group layer (not just the id), that can be retrieved via `.get(_group)`. Is this ok?

silvester-pari (approved):

PR #222, opened 2023-08-31
Tooltip enhancements

silvester-pari:
This PR introduces some enhancements for the eox-map-tooltip:
- it allows passing native OL parameters to the Overlay
- it adds some default styling of the tooltip component
- it adds positioning of the tooltip (top-left/right, bottom-left/right) depending on the mouse position within the map
- it adds the functionality to override the rendering of the feature properties.


RobertOrthofer (approved):
looks good to me

Issue #172, opened 2023-07-31
Add drawing of "box"


silvester-pari:
In some cases, it would be beneficial to be able to draw a "box" instead of a polygon (e.g. when selecting a bbox). There is a `createBox` functionality provided by OL, see https://openlayers.org/en/latest/examples/draw-shapes.html.

Once available in the `eox-map`, it would be used also in `eox-itemfilter` (spatial filtering, maybe as an option instead of polygon).

silvester-pari:
Implemented in https://github.com/EOX-A/EOxElements/pull/231.

Issue #149, opened 2023-07-17
Add "point" drawing functionality


silvester-pari:
The current implementation only supports drawing polygons. It would be nice to at least add "point" as option.

In previous apps we had the situation where a server API needed a polygon as input but the UI needed a point selection. We mitigated this by creating a buffer around a clicked point and emitting that. So the use cases could be:
- select a point on a map, emit a point
- select a point on a map, but emit a polygon which buffers the point

RobertOrthofer:
the `draw`-functionality of the `map`-element already submits a CustomEvent `drawend`, where additional information like a ready-to-use geojson is attached. We tried to move away from this concept, as it seems redundant to the normal event, that the ol-map emits. 
Maybe we should expose some methods to create these custom formats? Something that can easily create the needed output (geojson, buffered point, etc) with the information the callback of the ol-events provide (like a coordinate or a feature)

silvester-pari:
Point drawing available, buffer helper function now introduced in https://github.com/EOX-A/EOxElements/pull/231.

PR #231, opened 2023-09-01
Map/feature/draw box

RobertOrthofer:
This implements the feature to draw a "Box" (#172) by setting the `type` of the draw interaction to `Box`.

Additionally, this adds a solution to #149 exposing the extent function `buffer`. I am not certain what format the API in question expects, but using this as a starting-point we can send the buffer or single coordinates from the rectangle. Alternatively we can also expose a functionality to create a `GeoJSON` from the buffered extent.


silvester-pari (approved):
Thanks, works great!

PR #237, opened 2023-09-08
Remove map layers and groups

spectrachrome:
This pull request adds a delete button for layers and groups along with a Cypress test which checks if the entry was actually removed in both the map and the layer picker.

Screenshot 2023-09-18 at 11 25 37


silvester-pari (changes requested):
Thanks for this, please don't use the "real" eox-map in the tests, only a mocked map. This helps keeping these tests self-contained.

spectrachrome:
Button style has been implemented:

Screenshot 2023-09-11 at 15 43 55

The test case and mock map have been modified to test the delete functionality using the mock component instead, but it's still a work in progress.

PR #247, opened 2023-09-18
Itemfilter: autocomplete & inline mode

silvester-pari:
This PR introduces the new internal components `eox-itemfilter-autocomplete` and `eox-itemfilter-inline`

### Autocomplete
The `eox-itemfilter-autocomplete` component is an input field with multi select, whereas selections are displayed as (clearable) chips. This can be used to filter long lists of checkboxes (by default enabled if more than 10 filter options):

image

### Inline
The `eox-itemfilter-inline` component introduces a multi-step process to first select a filter, then selecting a filter value, by displaying the corresponding filters in a dropdown.
image

image

image


silvester-pari:
Superseded by https://github.com/EOX-A/EOxElements/pull/320

Issue #234, opened 2023-09-01
Add WMS/XYZ layer


silvester-pari:
Similar to the "optional layers" flow, the user could add external WMS or XYZ layers (or even more) to the map. Depends on #233 to create layers via JSON.

### Usecases
We have two usecases:
- `eox-layercontrol` in conjunction with the `eox-map` - here we can create layers using a JSON and e.g. `addOrUpdateLayer` method
- `eox-layercontrol` in conjunction with a vanilla OpenLayers map - in this case, we would need to create the layer directly in the layercontrol and then push it to the map. For this, we could either import OL modules, or import a `createLayer` helper function from `eox-map`

For teh second usecase, we need to investigate how much an OL import would affect the bundle size, and think of potential version conflicts between the OL version inside the layercontrol and the one used in a vanilla map. If it turns out to be too much of a hassle, we could (for now) opt for a check - if the layercontrol is attached to an `eox-map`, then offer the "add layer" functionality, otherwise not.

### Functionality
It would be great to have a dedicated button ([icon suggestion](https://pictogrammers.com/library/mdi/icon/layers-plus/)) somewhere in the layercontrol UI that opens the "add layer" UI - either inline, or with a popup etc. Inside this UI the user should be able to add a WMS/XYZ url including the placeholders for coordinates and/or parameters. In case the layercontrol is connected to an `eox-map`, it could also be nice to offer a JSON input for an eox-map style layer JSON.

Evolution of the bare functionality could include offering a form (based on `eox-jsonform`) for the parameters instead of adding just a link.

Further inspiration could be taken from e.g. https://viglino.github.io/ol-ext/examples/control/map.switcher.shop.html

PR #262, opened 2023-09-25
Always move a removed layer to the optional layer array

silvester-pari:
This PR introduces a change, that removal of a layer actually moves it to the optional layers array instead of removing it from the collection:
- visibility is set to `none`
- `layerControlOptional` is set to `true`,
- node in layer list is removed
- node in optional selection is added

### DIFF
--- a/elements/layercontrol/src/main.ts
+++ b/elements/layercontrol/src/main.ts
@@ -220,11 +220,7 @@ export class EOxLayerControl extends LitElement {
     };
     checkCollection(collection);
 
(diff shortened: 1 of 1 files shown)


silvester-pari:
I am not 100% sure about this, since with #234 we might introduce a way to add arbitrary layers to the layer control, and there it would make sense to really allow deleting a layer fully from a collection :thinking:

silvester-pari:
Superseded by https://github.com/EOX-A/EOxElements/pull/271

Issue #226, opened 2023-08-31
Programmatically select feature on map


StefanBrand:
For feature parity with the previous map element, we need a way to select a feature programmatically. I.e. the app starts and a feature is already highlighted from the beginning.

StefanBrand:
The current workflow is:
1. read feature ID from query params
2. ask backend for coordinates of feature
3. pan map to those coordinates
4. highlight feature

RobertOrthofer:
In the case of a `VectorTile`-Layer we cannot guarantee that the feature is available when the programmatic selection is initiated, so we would need to zoom to the location first. When working with slow tiles we could try to prevent the user from interacting with the map until everything is loaded and the feature is available.

I think we would need a list of IDs of features to be highlighted (if multi-select is a thing), as well as a single extent containing these features.

silvester-pari:
> The current workflow is:
> 
> 1. read feature ID from query params
> 2. ask backend for coordinates of feature
> 3. pan map to those coordinates
> 4. highlight feature

In my opinion, steps 1-3 would be application logic, and only step 4 would be a common functionality for the map element. So in the eox-map, we add functionality to highlight a feature **if** it is currently on the map, otherwise it could return an error/info that the feature is not currently on the map.

RobertOrthofer:
I would implement this as a method of an existing `select`-interaction, as this would handle the styling and management of selected features

PR #266, opened 2023-09-27
Map/feature/select by

RobertOrthofer:
This PR implements the map-part of the in #226 defined programmatic selection.

This includes:
 - refactoring of the selection to class structure
 - storing select-interactions in a dictionary
 - ability to remove select-interactions from the map
 - ability to set interactions active state via `selectInteraction.setActive()`
 - ability to highlight one or more features via `selectInteraction.highlightById([id])`

This programmatic highlighting is purely visual and does not fire a `select`-CustomEvent, as the manual selection does, as additional UI-actions are meant to be handled by the parent application.
Since the highlighting is purely visual, no additional tests are added.

### DIFF
--- a/elements/map/src/select.ts
+++ b/elements/map/src/select.ts
@@ -29,10 +29,11 @@ export class EoxSelectInteraction {
   options: SelectOptions;
   active: boolean;
   tooltip: HTMLElement;
-  selectedFid: string | number;
+  selectedFids: Array<string | number>;
   selectLayer: VectorTileLayer | VectorLayer<VectorSource>;
   selectStyleLayer: VectorTileLayer | VectorLayer<VectorSource>;
-  layerDefinition: any;
+  changeSourceListener: () => void;
+  removeListener: () => void;
 
   constructor(eoxMap: EOxMap, layerId: string, options: SelectOptions) {
     this.eoxMap = eoxMap;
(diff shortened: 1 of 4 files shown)


elements/map/src/select.ts
```diff
@@ -20,44 +20,189 @@ export type SelectOptions = Omit<
...
 
-export function addSelect(
-  EOxMap: EOxMap,
-  layerId: string,
-  options: SelectOptions
-) {
-  if (EOxMap.interactions[options.id]) {
-    throw Error(`Interaction with id: ${options.id} already exists.`);
-  }
+export class EoxSelectInteraction {
```
silvester-pari, line 26: Just a minor naming thing, we normally prefix all our classes with `EOx`, so in this case this could be `EOxSelectInteraction` if that's ok with you.

silvester-pari (changes requested):
Looks great in general, thanks! Regarding tests: would it be possible to programmatically select e.g. two features, and then check if the temporary layer has two features? Or would this need many changes? Not sure if the temporary layer is easily accessible.

elements/map/src/select.ts
RobertOrthofer: sure, I was waffling back and forth between the 2 naming options anyways

PR #349, opened 2023-10-19
chore: create commit hooks for eslint and prettier using `pre-commit`

spectrachrome:


silvester-pari:
Now that we have the working PR checks, I guess this is not necessarily needed any longer. Even if so, it would probably be done using a different approach (npm package). Closing this for now.

PR #369, opened 2023-10-25
chore: base structure of map stories

RobertOrthofer:
This PR adds stories of the `EOxMap` element. 
Right now, functionality that is set via methods or properties after the map is mounted (like selection, map sync, etc) are set via storybook [play functions](https://storybook.js.org/docs/react/writing-stories/play-function), as that is basically callback of the render function, if i understand correctly. While this works for single stories, these play functions are not called in the `documentation`-tab of storybook, and the code, which is the relevant part of the documentation, is also not displayed.

I'm not sure what the best solution to this is, or if this just displays some anti-pattern when designing web components. We can always duplicate the code snippets and display them in storybook inside the render function, but that does not fix the `documentation`-tab.
I have the feeling that we are supposed to use [args](https://storybook.js.org/docs/react/writing-stories/args#page-top) here, but I did not find the perfect solution.

### DIFF
--- a/elements/map/map.stories.js
+++ b/elements/map/map.stories.js
@@ -14,11 +14,371 @@ export const Primary = {
   render: (args) =>
     html`
       <eox-map
-        style="width: 400px; height: 300px;"
+        style="width: 100%; height: 300px;"
         zoom="${args.zoom}"
         center="[15, 48]"
         layers='[{"type":"Tile","source":{"type":"OSM"}}]'
       ></eox-map>
+    `,
+};
+
+export const VectorLayer = {
+  render: () =>
+    html` <style>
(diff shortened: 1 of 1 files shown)


silvester-pari (approved):
Thank you for this initial set of stories, which, while not yet perfect (play functions etc.), already give a first impression of what the map can do.
Probably the best would be to follow our concepts on how to define selects, controls etc. directly as a reactive property instead of exposing methods for this, but we can adjust the stories along the way when we introduce this.

The only thing I noticed which needs addressing soon is the STAC layer, which actually requires to register proj4 since it has a different projection. We could find a different example or handle this in a different way.

Issue #248, opened 2023-09-19
If a layer has minZoom/maxZoom set, change CSS color based on map current view zoom


lubojr:
If layer is currently not seen on the map view due to maxZoom/minZoom, this should be ideally reflected in the layercontrol by a different style/color for example.

santilland:
Cool idea! Maybe like a warning sign and when you hover over it, it shows a message? Something like... "This layer can't be shown on the current zoom level" ?

lubojr:
The official (vanilla) OL layer control used I think disabling of the input field + grey color, which sounds more mobile friendly to me, but at the same time disabling the field completely might be too limiting (if user wants to disable the layer before zooming towards it for whatever reason)

silvester-pari:
Implemented in https://github.com/EOX-A/EOxElements/commit/09f5288de6e86aab72f8f6cee0ba4cce337c7cfe

PR #370, opened 2023-10-25
Feature: Add layer zoom state

srijitcoder:
### What is this PR about (Based on #248)

- This PR allows helps us to view zoom level state based on `min` and `max` zoom
- The color change state only visible when `showZoomLayerState` is set inside layer properties.
  
- The background color of layer change to `#f0f5f9` and `opacity` to `0.3` (From -> TO)
   

### Demo - [Loom link here](https://www.loom.com/share/ec82e250b58e4d3ba7a85c98788e8132)
  <a href="https://www.loom.com/share/ec82e250b58e4d3ba7a85c98788e8132"></a>

### DIFF
--- a/elements/layercontrol/src/components/layer.js
+++ b/elements/layercontrol/src/components/layer.js
@@ -12,6 +12,7 @@ import { checkbox } from "../../../../utils/styles/checkbox";
 export class EOxLayerControlLayer extends LitElement {
   static properties = {
     layer: { attribute: false },
+    map: { attribute: false, state: true },
     titleProperty: { attribute: "title-property", type: String },
     tools: { attribute: false },
     unstyled: { type: Boolean },
@@ -26,6 +27,13 @@ export class EOxLayerControlLayer extends LitElement {
      * @see {@link https://openlayers.org/en/latest/apidoc/module-ol_layer_Layer-Layer.html}
      */
     this.layer = null;
+
+    /**
+     * The native OL map
+     * @type {import("ol").Map}
+     * @see {@link https://openlayers.org/en/latest/apidoc/module-ol_Map-Map.html}
+     */
+    this.map = null;
 
     /**
      * The layer title property
@@ -43,6 +51,15 @@ export class EOxLayerControlLayer extends LitElement {
     this.unstyled = false;
   }
 
+  updated() {
+    this.map.getView().on("change:resolution", () => {
+      this.requestUpdate();
+      this.dispatchEvent(
+        new CustomEvent("change:resolution", { bubbles: true, composed: true })
+      );
+    });
+  }
+
   createRenderRoot() {
     return this;
   }
@@ -56,7 +73,9 @@ export class EOxLayerControlLayer extends LitElement {
       ${when(
         this.layer,
         () => html`
-          <div class="layer ${this.layer.getVisible() ? "visible" : ""}">
+          <div class="layer ${this.layer.getVisible() ? "visible" : ""} ${
+            this.map.getView().getZoom() < this.layer.get('minZoom') || this.map.getView().getZoom() > this.layer.get('maxZoom') ? "zoom-state-invisible" : ""
+          }">
             <label
               class="${this.layer.get("layerControlDisable") ? "disabled" : ""}"
             >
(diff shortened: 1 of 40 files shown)


elements/layercontrol/src/components/layer.js
```diff
@@ -43,6 +52,21 @@ export class EOxLayerControlLayer extends LitElement {
     this.unstyled = false;
   }
 
+  /**
+   * Run zoom event to zoom level change
+   * and update DOM if `showZoomLayerState` present in layer property
+   */
+  updated() {
+    if (this.layer?.get("showZoomLayerState")) {
+      this.map.getView().on("change:resolution", () => {
```
silvester-pari, line 61: Registering the listener for `change:resolution` each time the element updates (`updated()`), the listeners actually stack and the more the element it uses, the more time this re-rendering triggers.

Adding a console logs in e.g. lines 61 and 64 shows this issue, which seems to slow down the website after zooming back and forth a bit and changing changing layer order:

image
image

silvester-pari (changes requested):
Thanks for this, it looks really promising, and we might be able to re-use a similar approach for other constraints as well, e.g. if a layer has a defined extent and the current map view is outside that extent. In that case the check could be something like `isLayerVisibleBasedOnZoomState && isLayerVisibleBasedOnExtent`. I have created https://github.com/EOX-A/EOxElements/issues/372 in order to track this.

If the stacking/loop of listeners as described in the comment is resolved, I think we can use this for the zoom for now and add the extent condition as described in the new issue at a later date.

elements/layercontrol/src/components/layer.js
srijitcoder: @silvester-pari 
I have resolved re-rendering and event stacking issue.
I have used `firstUpdated()` instead of `updated()` to stop stacking of event 

But it will still create event of layer which has `showZoomLayerState: true`. We can still use one event function by putting everything inside `layerList` and controlling it with a list of zoom layer states but then it won't make sense because the functionality should be part of `layer` not it's list.

silvester-pari (changes requested):
Thank you for the changes. This looks good, although I'd have three more suggestions for changing before merging:

- If possible, please undo the formatting changes in the markdown files. The CHANGELOG.md are created automatically from our release pipeline and I am not sure if this will case conflicts. We should somehow exclude the CHANGELOG.md from formatting alltogether, I added https://github.com/EOX-A/EOxElements/issues/401 in order to track this.
- Thanks for the example story, however I would suggest changing it a bit in order to make it more understandable. For example: only two layers (no group), WIND as background and regions as foreground, where regions has a `minZoom` of e.g. 5 and WIND has a `maxZoom` of e.g. 9. This would be an example of a real-world usecase, since vector layers (which slow down rendering) are usually hidden on lower zoom and raster layers (which get blurry or pixely) are usually hidden on higher zoom.
- instead of having `showZoomLayerState` on a layer level, I'd suggest having it as a configuration option for the entire layercontrol; my thought process is that if in some application the developer wants this functionality, it should be usually applied to all layers which have a minZoom and maxZoom. Having to declare this on every layer seems a bit verbose.

Also, it would be great if we could add a quick test case using [isVisible()](https://openlayers.org/en/latest/apidoc/module-ol_layer_Layer-Layer.html#isVisible).

srijitcoder:
> Thank you for the changes. This looks good, although I'd have three more suggestions for changing before merging:

I think all the points are straight forward. I will resolve these points first and then work on #371 changes.

srijitcoder:
@silvester-pari  All the points have been addressed in the second set of changes:

1. Reverted all changes in `CHANGELOG.md`.
2. Implemented example changes reflecting real scenarios - Included `minZoom: 2` for `Region` and `maxZoom: 9` for `WIND`.
3. Introduced `showZoomLayerState` to `eox-layercontrol`.
4. Incorporated test cases covering three scenarios:
  I. Default zoom level set to `1` (verifying `foo` is disabled).
  II. Zoom level adjusted to `3` (ensuring both layers are visible).
  III. Zoom level adjusted to `10` (verifying `bar` is disabled).

silvester-pari (approved):
Great contribution, thanks!

Issue #136, opened 2023-07-02
Add display of drawn features


silvester-pari:
After the user draws some features, they should be rendered as a list.
The user should also be able to select the features from the list (might need some additions in `eox-map`) or delete individual features. On feature select it would be nice if the map would zoom to that feature.

silvester-pari:
Implemented in https://github.com/EOX-A/EOxElements/pull/371

PR #371, opened 2023-10-26
Feature: Display of drawn feature

srijitcoder:
### What is this PR about (Based on #136)
- List of drawn features 
- Select individual feature from this list 
- Hover and highlight which feature is listed down 
- Select and zoom in to the selected feature 
- Delete individual feature 
- List gets updated as we draw 
- Enable it using `show-list` attribute

image

### Demo - [Loom link here](https://www.loom.com/share/2a307fdb45aa4b6e882e574516dfa7b1)

<a href="https://www.loom.com/share/2a307fdb45aa4b6e882e574516dfa7b1"></a>

### DIFF
--- a/elements/drawtools/src/components/list.js
+++ b/elements/drawtools/src/components/list.js
@@ -87,7 +87,7 @@ export class EOxDrawToolsList extends LitElement {
     // Else If selected index is greater than deleted feature index then changing index cursor by 1
     else if (this.selectedFeatureIndex > Number(index)) {
       this.selectedFeatureIndex = this.selectedFeatureIndex - 1;
-      this._handleSelectFeature(
+      this._handleFeatureSelectAndDeselect(
         this.selectedFeatureIndex,
         this.drawnFeatures[this.selectedFeatureIndex]
       );
@@ -96,15 +96,20 @@ export class EOxDrawToolsList extends LitElement {
   }
 
   /**
-   * Select a feature @param {Number} i
-   * Select a feature @param {import("ol").Feature} feature
+   * Select and Deselect feature from the list
+   *
+   * @param {Number} i
+   * @param {import("ol").Feature} feature
    */
-  _handleSelectFeature(i, feature) {
+  _handleFeatureSelectAndDeselect(i, feature) {
+    // Deselect selected feature
     if (this.selectedFeatureIndex === i) {
       const newExtent = this.drawLayer.getSource().getExtent();
       this.olMap.getView().fit(newExtent, { duration: 750 });
       this.selectedFeatureIndex = null;
-    } else {
+    }
+    // Select the clicked feature
+    else {
       if (this.selectedFeatureIndex !== null)
         this.drawnFeatures[this.selectedFeatureIndex].setStyle(
           getDefaultPolygonStyle()
@@ -144,7 +149,8 @@ export class EOxDrawToolsList extends LitElement {
               >
                 <div
                   class="list"
-                  @click="${() => this._handleSelectFeature(i, feature)}"
+                  @click="${() =>
+                    this._handleFeatureSelectAndDeselect(i, feature)}"
                 >
                   <span class="title">Feature #${i + 1}</span>
                   <button
--- a/elements/drawtools/src/components/list.js
+++ b/elements/drawtools/src/components/list.js
@@ -1,5 +1,8 @@
 import { LitElement, html, nothing } from "lit";
+import { keyed } from "lit/directives/keyed.js";
 import { styleEOX, styleListEOX } from "../style.eox";
+import * as ol from "ol/style.js";
+import { getDefaultPolygonStyle, getSelectedPolygonStyle } from "../helpers";
 
 /**
  * Display list of features
@@ -61,11 +64,20 @@ export class EOxDrawToolsList extends LitElement {
      * Render the element without additional styles
      */
     this.unstyled = false;
+
+    /**
+     * Index of selected feature
+     * @type Number || null
+     */
+    this.selectedFeatureIndex = null;
   }
 
   render() {
-    console.log(this.drawLayer);
+    console.log(this.drawLayer.getStyle());
     console.log(this.drawnFeatures[0]?.getProperties());
+    console.log(this.drawnFeatures[0]?.get("ol_uid"));
+    console.log(this.drawnFeatures[0]?.getGeometry().getExtent());
+    console.log(this.olMap.getView());
 
     return html`
       <style>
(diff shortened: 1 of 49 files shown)


elements/drawtools/src/helpers.js
```diff
@@ -0,0 +1,41 @@
+import * as ol from "ol/style.js";
```
silvester-pari, line 1: Until now, we haven't actually imported any OL modules in the drawtools, and we only had it as dev dependency for type checking.
Since OL supports [flat styles](https://openlayers.org/en/latest/apidoc/module-ol_style_flat.html), I think this could be easily replaced with a JSON style definition that doesn't need an actual module import.

elements/drawtools/src/style.eox.js
```diff
@@ -15,3 +15,80 @@ button.polygon:before {
...
+  }
+  label, span {
+    display: flex;
+    align-items: center;
+    cursor: pointer;
+    font-size: small;
+    flex-grow: 1;
+  }
+  .list .remove-icon::before {
+    content: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' fill='%23ff0000' viewBox='0 0 24 24'%3E%3Ctitle%3Edelete-outline%3C/title%3E%3Cpath d='M6,19A2,2 0 0,0 8,21H16A2,2 0 0,0 18,19V7H6V19M8,9H16V19H8V9M15.5,4L14.5,3H9.5L8.5,4H5V6H19V4H15.5Z' /%3E%3C/svg%3E");
```
silvester-pari, line 78: Here it would probably best to harmonize things and use the same icon as the `discard` button (see L11).

elements/drawtools/src/style.eox.js
```diff
@@ -15,3 +15,80 @@ button.polygon:before {
   content: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' fill='%23004170' viewBox='0 0 24 24'%3E%3Ctitle%3Eshape-polygon-plus%3C/title%3E%3Cpath d='M17,15.7V13H19V17L10,21L3,14L7,5H11V7H8.3L5.4,13.6L10.4,18.6L17,15.7M22,5V7H19V10H17V7H14V5H17V2H19V5H22Z' /%3E%3C/svg%3E");
 }
 `;
+
+export const styleListEOX = `
```
silvester-pari, line 19: I am wondering if there is a benefit to having this style as a separate export but within the same file. Maybe it would work to contain this style directly inside the `list.js`, but I don't have too strong of an opinion here.

elements/drawtools/src/components/list.js
```diff
@@ -0,0 +1,164 @@
...
+    }
+    this.requestUpdate();
+  }
+
+  /**
+   * Select a feature @param {Number} i
+   * Select a feature @param {import("ol").Feature} feature
+   */
+  _handleSelectFeature(i, feature) {
+    if (this.selectedFeatureIndex === i) return;
```
silvester-pari, line 107: Here we could actually allow the deselection of the feature, as there is currently no way to deselect (and reset the style of) a feature once it has been selected (other than selecting a different one).

elements/drawtools/src/components/list.js
```diff
@@ -0,0 +1,164 @@
...
+    if (this.selectedFeatureIndex !== null)
+      this.drawnFeatures[this.selectedFeatureIndex].setStyle(
+        getDefaultPolygonStyle()
+      );
+    feature.setStyle(getSelectedPolygonStyle());
+    this.olMap.getView().fit(feature.getGeometry().getExtent());
```
silvester-pari, line 113: Here it would be nice to add some small animation by providing some `options` to the fit, as currently it is hard for the user to see the context of where the map actually zoomed to.

elements/drawtools/src/helpers.js
srijitcoder: @silvester-pari 
We cannot use `FlatStyle` in `setStyle` of individual `features`.  We need to use `StyleLike`
So. should I add `ol` to dependency? 

Screenshot 2023-11-01 at 5 13 12 PM Screenshot 2023-11-01 at 5 13 41 PM

elements/drawtools/src/style.eox.js
srijitcoder: Done ✅

elements/drawtools/src/style.eox.js
srijitcoder: Done ✅

elements/drawtools/src/components/list.js
srijitcoder: Done ✅

elements/drawtools/src/components/list.js
srijitcoder: Done ✅

elements/drawtools/src/helpers.js
silvester-pari: You're correct, I wasn't aware that `FlatStyles` can only be applied to layers. One other option would be to set a property to the `Feature` and have a conditional styling on the layer level, for example:

```
// list.js
@mouseover=${() => {
  if (this.selectedFeatureIndex === i) return;
  feature.set("_highlighted", true);
}}
@mouseout=${() => {
  if (this.selectedFeatureIndex === i) return;
  feature.unset("_highlighted");
}}

// layer definition
{
  "type": "Vector","id": "draw","source": {"type": "Vector"},
  "style": [
    {
      "filter": ["get", "_highlighted"],
      "style": {
        "fill-color": "red"
      }
    },
    {
      "else": true,
      "style": {
        "fill-color": "blue"
      }
    }
  ]
},
```

Like this, the "highlighted" state is actually stored in the feature (which could come in handy for other use cases as well), but not the styling, so we would have a separation between features and styling.

If done like this, I'd suggest applying a default layer style to the vector layer if none is provided by the developer, but let the developer also define a custom feature highlighting style.

When using the drawtools in conjunction with `eox-map` we could also use the "programmatic selecting" functionality as introduced here: https://github.com/EOX-A/EOxElements/pull/266. This would require setting up an interaction on the map (e.g. a hover or a click interaction, useful for selecting features on the map directly), and then call `selectInteraction.highlightById([id])`.

Both solutions would be fine and wouldn't need the `ol` dependency. From my current understanding, the first solution would be more generic and would allow using the drawtools also with other OL maps, whereas the second solution would couple the drawtools closer to the `eox-map`, but would pave the way for further feature selection evolution.

silvester-pari (changes requested):
Also here, it would be great to not include the formatting changes, as explained in https://github.com/EOX-A/EOxElements/pull/370.

silvester-pari (approved):

PR #389, opened 2023-11-03
feat: remove the overlay when pointer leaves

### LINKED ISSUE #351: Detect `mouseleave` when `pointermove`-selection is active
Currently, tooltips stick to the side of the map when the mouse leaves the map element. This should be prevented by listeners on the map element.

RobertOrthofer:
This PR implements the in #351 documented issue of sticky tooltips in the hover-selection. 

closes #351


silvester-pari (approved):
Works great, thanks!

PR #390, opened 2023-11-05
fix: add styleOverride prop to stacinfo

lubojr:
inspired by https://github.com/EOX-A/EOxElements/pull/386 adding styleoverride to stacinfo

### DIFF
--- a/elements/stacinfo/src/main.ts
+++ b/elements/stacinfo/src/main.ts
@@ -47,7 +47,7 @@ export class EOxStacInfo extends LitElement {
   @property({ type: Array })
   footer: Array<string> = [];
 
-  @property()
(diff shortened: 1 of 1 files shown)


elements/stacinfo/src/main.ts
```diff
@@ -47,6 +47,9 @@ export class EOxStacInfo extends LitElement {
   @property({ type: Array })
   footer: Array<string> = [];
 
+  @property()
```
silvester-pari, line 50: For completeness, please add `@property({ attribute: "style-override" })` so it also can be set via HTML attribute `style-override` in a vanilla use case!

silvester-pari (changes requested):

silvester-pari (approved):

PR #391, opened 2023-11-06
fix: optional list update after layer removal

### LINKED ISSUE #342: not all removed layers are shown instantly on the optional layers drop down
On the removal of layers only the first removed layer is shown instantly on the optional layers drop down, and the layers removed afterwards are shown on the next update.

A-Behairi:
this PR closes #342; `eox-layercontrol-optional-list` gets updated after any layer removal.


silvester-pari (approved):
Thanks, LGTM!

PR #413, opened 2023-11-09
feat: add reset for individual filters

A-Behairi:
this introduces a reset functionality for every filter type in `eox-itemfilter`. When a user chooses to assign values to the filter, a reset icon appears next to the filter title and allows to reset the properties of that specific filter.


silvester-pari (approved):
Thank you, works great! I chose a simpler icon (X) for now, I feel like this is easier on the eyes. I also tried to harmonize it with the "reset all" button.

image

I think we can merge this now, we can still keep looking for better icons in the future.

Issue #453, opened 2023-11-22
Refactoring `drawtools`


srijitcoder:
- **Atomic Structuring:** This approach facilitates easy management and readability of small chunks of code, allowing us to separate them based on their functionality.

  - `/src/methods`
      - All method and handler functions inside components belong in this folder.
      - Segregating these functions reduces the lines of code within components.
      - We'll place them in individual files like `method1.js`, `method2.js`, and combine them inside `index.js` (to be imported as:

    ```javascript
    import { method1, method2 } from "/methods";
    ```

  - `/src/helpers`
      - Currently, there's only one `helpers.js` file containing multiple functions.
      - We'll separate them into individual files like `helper1.js`, `helper2.js`, combining them in `index.js` (to be imported as:

    ```javascript
    import { helper1, helper2 } from "/helpers";
    ```

  - `/src/components`
      - All components will reside within this folder.

  - `/src/style`
      - All style files will be stored in this folder.

  - `main.js` - Entry point for the library.
  - `enums.js`
      - Contains keywords or objects repeatedly used in our code, placed in uppercase. For example:

    ```javascript
    const DRAW_EVENTS = {
       click: "draw-click",
       hover: "draw-hover"
    };

    DRAW_EVENT.click;
    ```

  - `/stories`
      - All stories will be placed in the `/stories` folder.
      - To prevent unwieldy files, stories will be split into smaller, atomic-structured individual files such as `one.stories.js`, `two.stories.js`.

  - `/test`
      - Test cases within `general.cy.js` will be separately imported from the `/test/cases` folder:

    ```javascript
    import { case1, case2 } from "/cases";

    it("Test case 1", () => case1());
    it("Test case 2", () => case2());
    ```

- Ensuring comprehensive JavaScript type checking.
- Simplification of complex functions.
- Including adequate comments as required.
- Identifying and rectifying any re-rendering or event function triggers.
- Adding more test cases.
- @silvester-pari Can we experiment with Tailwind CSS? (It's an excellent styling library that reduces the need for extensive CSS).

```
- jsonform
- layercontrol
- drawtools
    - src
        - style
            - style.eox.js
            - style.js
        - helpers
            - index.js
            - helper1.js
            - helper2.js
        - methods
            - index.js
            - method1.js
            - method2.js 
        - components
            - index.js
            - control.js
            - list.js
        - main.js
        - enum.js
    - stories
        - one.stories.js
        - two.stories.js
        - config.js
    - test
        - cases
[...]

PR #460, opened 2023-11-27
Enhancement: `drawtools` refactoring

srijitcoder:
### Important Changes 
- Segregated all the functions/methods into separate files to improve readability.
- Component-level methods are now organized within `/src/methods/draw`, `/src/methods/list`, and `/src/methods/controller`.
- Introduced `enums` for centralizing and managing constants.
- Stories have been segregated into distinct files and consolidated into a single file using `index.js`.
- Portions of the code were rewritten to enhance readability and reduce complexity.
- Extensive comments have been added for better comprehension.
- Test cases are now separated into individual files and consolidated using `index.js`.
- Efforts were made to ensure that component-level code remains under 180 lines and functions under 50 lines.

### Not in this PR 
- Styling refactoring: This will be addressed separately after researching and potentially using `tailwind`.

### For further details, 
- Refer to issue #453.

### DIFF
--- a/elements/drawtools/src/methods/draw/emit-drawn-features.js
+++ b/elements/drawtools/src/methods/draw/emit-drawn-features.js
@@ -2,20 +2,17 @@
  * Emits drawn features after a timeout to allow updating drawn features.
  *
  * @param {import("../../main").EOxDrawTools} EoxDrawTool - The drawing tool instance.
+ * @param {Function} drawUpdateEvent - event to be triggered after drawFeature is updated
  */
-const emitDrawnFeaturesMethod = (EoxDrawTool) => {
+const emitDrawnFeaturesMethod = (EoxDrawTool, drawUpdateEvent) => {
   // Function to emit features after a timeout (ensures update)
   const emitFeatures = () => {
     // Update drawnFeatures with features from drawLayer's source
     EoxDrawTool.drawnFeatures = EoxDrawTool.drawLayer.getSource().getFeatures();
     EoxDrawTool.requestUpdate();
 
-    // Dispatch a 'drawupdate' event with drawn features as detail
-    EoxDrawTool.dispatchEvent(
-      new CustomEvent("drawupdate", {
-        detail: EoxDrawTool.drawnFeatures,
-      })
-    );
+    // Triggering `drawupdate` event after drawFeature is updated
+    drawUpdateEvent();
   };
 
   // Emit features after a timeout (ensures update)
--- a/elements/drawtools/src/components/controller.js
+++ b/elements/drawtools/src/components/controller.js
@@ -62,12 +62,12 @@ export class EOxDrawToolsController extends LitElement {
   /**
    * Updates button states based on conditions
    */
-  updateButtonStates = () => {
+  updateButtonStates() {
     const { drawDisabled, discardDisabled } = updateButtonStatesMethod(this);
 
     this.#drawDisabled = drawDisabled;
     this.#discardDisabled = discardDisabled;
-  };
+  }
 
   render() {
     this.updateButtonStates();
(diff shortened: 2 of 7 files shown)


elements/drawtools/src/enums/draw.js
```diff
@@ -0,0 +1 @@
+export const TRANSITION_DURATION = { duration: 750 };
```
silvester-pari, line 1: The naming might be a bit confusing here, since these are actually the fit options (as described in the [OL docs](https://openlayers.org/en/latest/apidoc/module-ol_View-View.html#fit), so maybe `TRANSITION_OPTIONS` or `FIT_OPTIONS` could be better?

elements/drawtools/src/methods/controller/update-button-states.js
```diff
@@ -0,0 +1,28 @@
...
+const updateButtonStatesMethod = (EoxDrawToolController) => {
+  // Destructure relevant properties from the controller
+  const { multipleFeatures, drawnFeatures, currentlyDrawing } =
+    EoxDrawToolController;
+
+  // Determine draw button's disabled state based on conditions
+  const drawDisabled =
+    (!multipleFeatures && drawnFeatures?.length > 0) ||
+    currentlyDrawing ||
+    nothing;
```
silvester-pari, line 18: Is there a concrete reason for returning a lit `nothing` here (and L22)? This adds a dependency to this otherwise vanilla JS module.
Could we use something like `?disabled="${this.#discardDisabled || nothing}"` in the controller.js instead?

elements/drawtools/src/methods/draw/emit-drawn-features.js
```diff
@@ -0,0 +1,25 @@
...
+ */
+const emitDrawnFeaturesMethod = (EoxDrawTool) => {
+  // Function to emit features after a timeout (ensures update)
+  const emitFeatures = () => {
+    // Update drawnFeatures with features from drawLayer's source
+    EoxDrawTool.drawnFeatures = EoxDrawTool.drawLayer.getSource().getFeatures();
+    EoxDrawTool.requestUpdate();
+
+    // Dispatch a 'drawupdate' event with drawn features as detail
+    EoxDrawTool.dispatchEvent(
```
silvester-pari, line 14: There seems to be some issue with our custom element parser, as it doesn't recognize the emitted event in this case. If you compare with e.g. `drawtools`, it does if in the main.js file it directly does something like
```
  #emitDrawnFeatures() {
    ...
    this.dispatchEvent(
      new CustomEvent("drawupdate", {
        detail: EoxDrawTool.drawnFeatures,
      })
    );
  }
```
This could be mitigated by moving the dispatch to main.js, or maybe there is some other solution. The ideal outcome would show the event in the docs:

image

elements/drawtools/src/enums/draw.js
srijitcoder: Yeah this make sense. I will change it to `FIT_OPTION`, that will go much similar to OL Docs.

elements/drawtools/src/methods/controller/update-button-states.js
srijitcoder: Yeah you are right, I shouldn't have imported `nothing` inside a method file which is contradicting. `nothing` should only be part of components because lit it associated with that.

I will add it to `controller.js`

elements/drawtools/src/methods/draw/emit-drawn-features.js
srijitcoder: Thanks for noticing this out. I will try to find an optimal work around for this.

elements/drawtools/src/methods/controller/update-button-states.js
srijitcoder: Done ✅

elements/drawtools/src/methods/draw/emit-drawn-features.js
srijitcoder: Done ✅

PR #491, opened 2023-12-04
fix: sorting with optional & hidden layers

A-Behairi:
This solves the following issues for `eox-layercontrol-layer-list`:
1. sorting lists that contain layers that are not displayed.
2. sets `idProperty` & `titleProperty` if not assigned.


silvester-pari (approved):
Thank you!

PR #515, opened 2023-12-12
feat!: extract advanced layers & sources into separate bundle

### LINKED ISSUE #495: Improve bundle size using optional layer types/sources
Currently, the `eox-map` bundle is quite big as it imports all layers, sources and formats, plus ol-stac as well as proj4. We could offer an "all-inclusive" bundle fro prototyping, but for production use we should somehow offer a way to only include those layer/source/format types needed.

image

silvester-pari:
This PR splits up OpenLayers `formats`, `layers` and `sources` to include only the most basic ones, while advanced ones can be added as a separate plugin. This is in order to reduce the bundle size for the most common use cases.

## Before
image
image

## After
image
image

Included in the base bundle:
- Formats: `GeoJSON`, `MVT`
- Layers: `Group`, `Image`, `Tile`, `Vector`, `VectorTile`,
- Sources: `ImageWMS`, `OSM`, `Tile`, `TileWMS`, `Vector`, `VectorTile`, `WMTS`, `XYZ`

## Usage
In order to use all formats, layers and sources provided by OpenLayers, import the plugin as well:
```
import "@eox/map/dist/eox-map-advanced-layers-and-sources.js"
import "@eox/map/dist/eox-map.js"
<eox-map [... GeoTIFF sources, STAC layers etc ...]></eox-map>
```

If the developer forgets to import the bundle but tries to use an advanced format, layer or source, a console error is thrown:
image

Closes #495

### DIFF
--- a/elements/map/src/plugins/advancedLayersAndSources/layers.ts
+++ b/elements/map/src/plugins/advancedLayersAndSources/layers.ts
@@ -1,0 +1,11 @@
+import * as olLayers from "ol/layer";
+import STAC from "ol-stac";
+import { register } from "ol/proj/proj4.js";
+import proj4 from "proj4";
+
+register(proj4); // required to support source reprojection
+
+window.eoxMapAdvancedOlLayers = {
+  ...olLayers,
+  STAC,
+};
--- a/.eslintrc
+++ b/.eslintrc
@@ -9,6 +9,7 @@
   "plugins": ["cypress", "@typescript-eslint"],
   "root": true,
   "rules": {
-    "@typescript-eslint/ban-ts-comment": "warn"
+    "@typescript-eslint/ban-ts-comment": "warn",
+    "@typescript-eslint/no-explicit-any": "warn"
   }
 }
--- a/elements/map/README.md
+++ b/elements/map/README.md
@@ -16,19 +16,11 @@ import "@eox/map"
 <eox-map></eox-map>
 ```
 
-// TODO: add documentation
+Please refer to the [map docs](https://eox-a.github.io/EOxElements/?path=/docs/elements-eox-map--docs) for more details.
 
-## Contribute
+## Development
 
-```
-npm watch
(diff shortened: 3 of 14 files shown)


elements/map/src/plugins/advancedLayersAndSources/layers.ts
```diff
@@ -0,0 +1,11 @@
+import * as olLayers from "ol/layer";
+import STAC from "ol-stac";
+import { register } from "ol/proj/proj4.js";
+import proj4 from "proj4";
+
+register(proj4); // required to support source reprojection
```
StefanBrand, line 6: It seems that reprojection is only supported when the advanced plugin is imported. If this is true, we should document it in the README as well.

elements/map/src/plugins/advancedLayersAndSources/layers.ts
```diff
@@ -0,0 +1,11 @@
...
+
+window.eoxMapAdvancedOlLayers = {
+  ...olLayers,
+  STAC,
```
StefanBrand, line 10: The README currently states "In order to use all formats, layers and sources provided by OpenLayers, import the plugin as well". Since STAC is a plugin of it's own (and not provided directly by `ol`), it should also be documented that it is included in our advanced plugin.

StefanBrand (approved):
Interesting technique. I'm curious to find out if this directly affects the bundle size of the AgriApp, one of our downstream applications.

elements/map/src/plugins/advancedLayersAndSources/layers.ts
silvester-pari: Thank you, added to description!

elements/map/src/plugins/advancedLayersAndSources/layers.ts
silvester-pari: Thank you, added to description!

Issue #416, opened 2023-11-10
Allow custom ordering of Filters and Filtered Items in ItemFilter component


lubojr:
Investigate allowing a configuration of "custom ordering" of filtered items and filters themselves.

For example by direct matching against of ids to configured order and fallback to alphabetical if not matched?

Originating from https://github.com/eurodatacube/eodash/issues/2363#issuecomment-1797901979

**Edit 14.11.: It was confirmed that eodash integration only needs custom ordering of Filters, not the Results themselves.**

PR #526, opened 2023-12-14
Feat: Custom ordering for select and multi select filters

A-Behairi:
Introducing custom ordering function `sort` as a  property in `type: select` and `type: multiselect` filter objects. Closing #416


silvester-pari (approved):
Looks great, thanks!

Issue #325, opened 2023-10-18
Generic way of adding style overrides


silvester-pari:
Each element needs some way to override the styling; some concepts are already present in the itemfilter, but no default has been developed yet.

Related to the question of shadow vs. non-shadow.

silvester-pari:
Interesting summary here: https://nolanlawson.com/2021/01/03/options-for-styling-web-components/

lubojr:
Theming done via style variables in https://github.com/EOX-A/EOxElements/pull/616

PR #536, opened 2023-12-18
Fix: Itemfilter StyleOverride access to selectionlist

A-Behairi:
Introducing `noShadow` property in `eox-selectionlist` that enables it's rendering without a shadow root and giving access for `styleOverride` property of the `eox-itemfilter` to reach `eox-selectionlist` in primary (not inline) mode.

### DIFF
--- a/elements/itemfilter/src/filters/multiselect.ts
+++ b/elements/itemfilter/src/filters/multiselect.ts
@@ -87,7 +87,7 @@ export class EOxItemFilterMultiselect extends LitElement {
           `,
           () => html`
             <eox-selectionlist
-              ?noShadow=${!this.inline}
+              .noShadow=${!this.inline}
               multiple
               .items=${this._getItems()}
               .selectedItems=${this._getSelectedItems()}
--- a/elements/itemfilter/src/filters/multiselect.ts
+++ b/elements/itemfilter/src/filters/multiselect.ts
@@ -87,6 +87,7 @@ export class EOxItemFilterMultiselect extends LitElement {
           `,
           () => html`
             <eox-selectionlist
+              ?no-shadow=${!this.inline}
               multiple
               .items=${this._getItems()}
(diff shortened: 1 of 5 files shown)


lubojr:
Just to double check - this PR is going to fix the `styleOverride` for ItemFilter not working at the moment?

Storybook https://eox-a.github.io/EOxElements/?path=/docs/elements-eox-itemfilter--docs for ItemFilter does not interact anyhow to the changes of the `styleOverride` property or is the behavior inside storybook just a "nonreactivity" issue of the storybook itself?

silvester-pari:
> Just to double check - this PR is going to fix the `styleOverride` for ItemFilter not working at the moment?
> 
> Storybook https://eox-a.github.io/EOxElements/?path=/docs/elements-eox-itemfilter--docs for ItemFilter does not interact anyhow to the changes of the `styleOverride` property or is the behavior inside storybook just a "nonreactivity" issue of the storybook itself? 
>  

Yes and no - this PR fixes some styleOverride issues for the sub-component `selectionlist`, which currently isn't "reachable" for the override since it's located in a separate shadow root.

The non-reacting of the storybook example might be an other issue, could be that the styleOverride property is not yet reactive, means doesn't react on changes, only uses initial value.

Before investing too much effort into this, I'd like to come up with a good generic way for style overrides, see https://github.com/EOX-A/EOxElements/issues/325

lubojr:
OK. My concrete use case why I asked was that since ItemFilter 0.9.0 -> 0.12.0 the way how styleOverride works has changed (if I understand it correctly from filterOptions.styleOverride to property called styleOverride) but also the structure of HTML elements which compose the ItemFilter single select changed. Therefore current overrides for GTIF Domain images do not work anymore and I am struggling to get it back working in a branch https://github.com/eurodatacube/eodash/pull/2427.

Issue #548, opened 2023-12-22
Cannot pass custom tooltip


StefanBrand:
Since we `stringify` the layers property, it is no longer possible to pass in a HTML element as a tooltip (by setting `options.overlay.element` of an interaction).

https://github.com/EOX-A/EOxElements/blob/00d7af5efa6bce88912c701e8dd777b9f8c794ec/elements/map/main.ts#L114

It results in a [TypeError: cyclic object value - JavaScript | MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Cyclic_object_value).

RobertOrthofer:
since we were already discussing special features that will need need more complex data types, maybe this is the time to look into a simple diff-package (e.g. [deep-object-diff](https://www.npmjs.com/package/deep-object-diff)) to handle the change detection?
There also was the idea of keeping the map-definitions simple, so we could share map definitions via copy-paste or via api-calls somewhere in the future. IMO this can still be possible, as long as we keep all features that require complex data types optional. 
Another thing to consider is that change detection for complex data types will never be possible, as `{} !== {}`

PR #555, opened 2024-01-08
fix: check for tooltip element in shadowroot

### LINKED ISSUE #549: Setting layers multiple times destroys tooltip functionality
I'm setting `layers` prop with interactions multiple times, after each relevant user interaction. Tooltip is defined with `eox-map-tooltip` element. The first time the tooltip shows, but subsequently, no tooltip is shown and the tooltip has even vanished from the DOM.

Possibly, this line does not work when setting interactions multiple times:

https://github.com/EOX-A/EOxElements/blob/00d7af5efa6bce88912c701e8dd777b9f8c794ec/elements/map/src/select.ts#L53

I've tried with, but I was not successful so far. My insight is limited, however.

```js
    this.tooltip =
      this.eoxMap.querySelector("eox-map-tooltip") ||
      this.eoxMap.shadowRoot.querySelector("eox-map-tooltip") ||
      options.overlay?.element;
```

silvester-pari:
This PR adds an additional check to find the tooltip element inside the map shadow root and fixes #549. The reason for this is that OpenLayers "hijacks" the DOM element and pulls it into its DOM, changing the location from the slot (outside the shadow root) into its own div (inside the shadow root).
This fix does not consider the overlay element option, as this is tracked in a separate issue (#548).

### DIFF
--- a/elements/map/map.stories.js
+++ b/elements/map/map.stories.js
@@ -300,6 +300,7 @@ export const Tooltip = {
         interactions: [
           {
             type: "select",
+            showTooltip: true,
             options: {
               id: "selectInteraction",
               condition: "pointermove",
@@ -360,6 +361,7 @@ export const TooltipWithPropertyTransform = {
             options: {
               id: "selectInteraction",
               condition: "pointermove",
+              tooltipId: 'hoverTooltip',
               style: {
                 "stroke-color": "red",
                 "stroke-width": 3,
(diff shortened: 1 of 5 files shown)


StefanBrand:
This PR does not fix the issue. I can still reproduce it in the application.

StefanBrand:
Tooltips work for multiple layers now. However, we also get tooltips for `singleclick` interactions:

[Peek 2024-01-12 10-50.webm](https://github.com/EOX-A/EOxElements/assets/8948865/3614c304-4eb0-4de4-bfe1-2290d58a48d5)

StefanBrand:
Workaround is to remove the unnecessary overlays on `loadend`:

```
      olMap
        .getOverlays()
        .getArray()
        .filter((overlay) => overlay.options.element.id.includes(CLICK_INTERACTION_SUFFIX))
        .forEach((overlay) => olMap.removeOverlay(overlay));
```

However the property transform seems to be broken. :grimacing:

silvester-pari:
Superseded by https://github.com/EOX-A/EOxElements/pull/571.

PR #563, opened 2024-01-10
Fix: Layercontrol external API test fail

srijitcoder:
### Context
- Final Solution for Failing API Requests in Cypress Tests

### What Does This PR Do?
- Implemented an interceptor to mock API requests with a dummy foo response.
- Rewrote the code to enhance asynchronous compatibility.
- Resolved input typing issues during browser interruptions by using force: true.
- Repeatedly ran tests locally and conducted 10 test iterations on GA.

image


silvester-pari (approved):
Very nice, thanks

PR #569, opened 2024-01-11
Feature: Disable/Enable scroll on `EOxMap` using `prevenScroll ` prop

srijitcoder:
### Context -
- Feature to disable/enable `scroll` on `EOxMap`

#### We can prevent scrolling with the help of `prevenScroll` property and it will affect two major interactions -
- `MouseWheelZoom`: Will be disabled using `platformModifierKeyOnly` condition 
- `DragPan`: Will be disabled only on `tab/mobile` devices and `platformModifierKeyOnly` condition

### Usage - 

```html
<eox-map
  .preventScroll=${true}
></eox-map>

or 

<eox-map
  .config=${{
    ...
    preventScroll: true,
  }}
></eox-map>

or

<eox-map
  prevent-scroll
></eox-map>
```


RobertOrthofer:
If this is only used for disabling interactions while scrolling, you can probably catch the corresponding event and then disable all interactions with `eoxMap.map.getInteractions().forEach(i => i.setActive(false))`.

If i see correctly, there is no way to re-activate the default interactions via the `disableInteractions `-prop, right? 

In my opinion, the root-level of the `ConfigObject` if the correct place to do this, but having the prop called `disableInteractions` feels somehow unusual... Maybe we could just call it `interactions` (or maybe `viewInteractions` or something like that, since we have "interactions" already as property of the layers) and set the default value to an array containing all our default interactions, like 

`[
      "DragRotate",
      "DoubleClickZoom",
      "DragPan",
      "PinchRotate",
      "PinchZoom",
      "KeyboardPan",
      "KeyboardZoom",
      "MouseWheelZoom",
      "DragZoom"
    ]`
    
and then we could handle changes to the array in a `setter`-function, similar to the `controls`, doing something like 
`allInteractions.forEach(interaction => setInteractionActive(newInteractions.includes(interaction)))`

elements/map/src/interaction.ts
```diff
@@ -0,0 +1,36 @@
...
+  disableInteractions: InteractionsType,
+  olMap: Map
+) {
+  olMap.getInteractions().forEach(function (interaction) {
+    disableInteractions.forEach(
+      (type) =>
+        interaction instanceof interactions[type] &&
+        interaction.setActive(false)
+    );
+  });
```
srijitcoder, line 35: > If this is only used for disabling interactions while scrolling, you can probably catch the corresponding event and then disable all interactions with `eoxMap.map.getInteractions().forEach(i => i.setActive(false))`.

Currently, I am using the same logic of `setInteraction()` by using `olMap` i.e. `EOx.map`.

srijitcoder (commented):
> If i see correctly, there is no way to re-activate the default interactions via the `disableInteractions `-prop, right? 

Yeah right now there is no way to revert it back to `default` interactions.

---

> In my opinion, the root-level of the `ConfigObject` if the correct place to do this, but having the prop called `disableInteractions` feels somehow unusual... Maybe we could just call it `interactions` (or maybe `viewInteractions` or something like that, since we have "interactions" already as property of the layers) and set the default value to an array containing all our default interactions, like 

We can rename it to something other than `disableInteractions` but as per you said we need to pass all the interactions list each time we need to disable any single `interactions`. For example, if I need to disable just `MouseWheelZoom`, then each time I need to pass this long list without `MouseWheelZoom`. Will that be fine?

silvester-pari:
I am not sure if we really want to just disable interactions (can't think of any usecase), or rather add conditions. See this example: https://openlayers.org/en/latest/examples/two-finger-pan-scroll.html
It does exactly what we need, add conditions on touch and desktop to prevent accidental scrolling, but enables scrolling by using a modifier.

If we add a property like e.g. "preventScroll" in the config object then we could add this functionality. Problem is it wouldn't be reactive, as I haven't yet found a way to enable/disable this after the Map was initialized.

silvester-pari (approved):

PR #571, opened 2024-01-15
fix: check if overlay was already instantiated before creating a new one

silvester-pari:
This PR goes in a different direction than https://github.com/EOX-A/EOxElements/pull/555: when the select interaction is initialized, first check if an eox-map-tooltip was already instantiated on the map, if so, reuse that overlay, if not, create a new one.
This should fix the issue with not finding the DOM element, but means only one tooltip can exist per map.

### DIFF
--- a/elements/map/map.stories.js
+++ b/elements/map/map.stories.js
@@ -285,7 +285,7 @@ export const ClickSelect = {
  * ```
  * This renders a list of all feature properties of the currently selected feature.
  * Note that if multiple interactions are registered (e.g. `pointermove` and `singleclick`),
- * only the first one will trigger the tooltip.
+ * the `pointermove` interaction will have higher priority for the tooltip.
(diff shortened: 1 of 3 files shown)


RobertOrthofer (approved):
While this does still seem a little hacky to me, i like the approach of going back to having a single tooltip element  for the entire map. 
This disregards tooltips for click selection, which is ok I suppose for the current use cases and the current API.

silvester-pari:
> While this does still seem a little hacky to me, i like the approach of going back to having a single tooltip element for the entire map. This disregards tooltips for click selection, which is ok I suppose for the current use cases and the current API.

Thanks for your review. While this indeed "breaks" the functionality to have e.g. different tooltips on hover and on click, the basic tooltip functionality should be more robust without introducing breaking changes to the API.

PR #588, opened 2024-01-16
fix: make sure a fallback layer id is rendered in time in order to support sorting

### LINKED ISSUE #584: Sorting removes layer in certain cases
As described in https://github.com/EOX-A/EOxElements/issues/580#issuecomment-1893799785, sorting sometims removes a layer completely:

https://github.com/EOX-A/EOxElements/assets/19269667/271f2040-dc36-4add-9e46-9f553657c9aa

Can e.g. be experienced at https://eox-a.github.io/EOxElements/?path=/story/elements-eox-layercontrol--exclusive-layers

silvester-pari:
This PR adds an additional `requestUpdate()` after the initial `propertyCheck()`, to force the DOM to re-render. The reason for this is that SortableJS needs the layer `id` present in order to work correctly.

The `requestUpdate()` is only called in case it is needed, e.g. if the `id` or `title` were changed by the `propertyCheck()`.

Fixes #584.

### DIFF
--- a/elements/layercontrol/src/helpers/check-properties.js
+++ b/elements/layercontrol/src/helpers/check-properties.js
@@ -5,20 +5,37 @@
  * @param {import("ol").Collection<import("ol/layer").Layer | import("ol/layer").Group>} collection - The collection of layers to be checked.
  * @param {string} idProperty - The property used as the ID.
  * @param {string} titleProperty - The property used as the title.
+ * @param {import("lit").LitElement} that - The LitElement instance.
  */
 //
-export default function checkProperties(collection, idProperty, titleProperty) {
+export default function checkProperties(
+  collection,
+  idProperty,
+  titleProperty,
+  that
+) {
   const layerArray = collection.getArray(); // Get an array of layers from the collection.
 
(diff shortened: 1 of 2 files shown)


elements/layercontrol/src/helpers/check-properties.js
```diff
@@ -5,20 +5,37 @@
...
+ * @param {import("lit").LitElement} that - The LitElement instance.
  */
 //
-export default function checkProperties(collection, idProperty, titleProperty) {
+export default function checkProperties(
+  collection,
+  idProperty,
+  titleProperty,
+  that
+) {
```
srijitcoder, line 16: Yeah, this PR looks good to me because `checkProperties()` is a part of `/helpers`, and in this case, we don't define the **actual instance** name like `EOxLayerControlLayerList`. Therefore, `that` seems to be the best property name, as this function can be reused in other components.

Additionally, according to the logic, it also makes sense to re-render the component if the `id` or `title` changes using `requestUpdateNecessary`.

srijitcoder (approved):

PR #590, opened 2024-01-17
chore: decrease bundle size of jsonform

StefanBrand:
Instead of importing the bundle, import the module, which allows tree-shaking.

Ref. https://github.com/json-editor/json-editor/issues/1470

## Before

```
dist/eox-jsonform.js  828.19 kB │ gzip: 144.86 kB
dist/eox-jsonform.umd.cjs  596.59 kB │ gzip: 127.46 kB
```

## After

```
dist/eox-jsonform.js  421.60 kB │ gzip: 86.46 kB
dist/eox-jsonform.umd.cjs  328.16 kB │ gzip: 74.71 kB
```


silvester-pari (approved):
Thank you for looking into this!

PR #604, opened 2024-01-25
feat: add `layerControlHideGroup` property option

silvester-pari:
## Implemented changes
This PR adds a layer property called `layerControlHideGroup` which allows rendering individual groups without their children, i.e. like a "normal layer".
The layer-type symbol is still shown as group to indicate that it is actually a group, but not rendered:

## Screenshots
image
---
image

- [x] All checks have passed


silvester-pari:
I am unsure about two things:
- the naming: if anyone has a better name than "hidden group" I'd gladly change it
- if we should display the group icon in the layer or not. It might be confusing, but I don't know which other icon to show there instead (the contents could be raster, vector, etc.)

srijitcoder:
### @silvester-pari, the name sounds okay to me. Maybe we can identify it using a tag and some opacity - 

image

silvester-pari:
> ### @silvester-pari, the name sounds okay to me. Maybe we can identify it using a tag and some opacity -

Thanks for the feedback, but it is not really about the group itself being hidden, but the layers inside the group are hidden, so only the group-level is shown (not possible to expand layers inside group). Maybe the name is confusing after all.

StefanBrand:
Maybe this should rather be a `hideLayer` -> individually hide layers. When group has only hidden layers, don't make it expandable.

silvester-pari:
> Maybe this should rather be a `hideLayer` -> individually hide layers. When group has only hidden layers, don't make it expandable.

That's a good idea, this functionality already exists, we'd just need to add a check how many layers inside a group are hidden. I wonder if this is complicated, I'll give it a try.

silvester-pari:
> > Maybe this should rather be a hideLayer -> individually hide layers. When group has only hidden layers, don't make it expandable.

> That's a good idea, this functionality already exists, we'd just need to add a check how many layers inside a group are hidden. I wonder if this is complicated, I'll give it a try.

Opened another PR that impolements just this: https://github.com/EOX-A/EOxElements/pull/651

PR #610, opened 2024-01-31
feat: add a `currentTime` property to get and set time programmatically

### LINKED ISSUE #608: Option to set the selected date programatically in TimeControl
Currently there is no option in the TimeControl element to set the selected date programatically, it selects the first element of the array specified in the `.animation_values` property. For the Agri-App implementations, this feature would be needed, so that when we select another date in different widgets, the TimeControl element also updates accordingly.

spectrachrome:
## Implemented changes

This pull request introduces a new `currentTime` property within the time control, which gets the currently selected time string of an `eox-timecontrol` element, for example like this:

```javascript
// Set the current time
timeControl.currentTime = '2022-12-26'

// '2022-12-26'
console.log(timeControl.currentTime)
```

Closes #608 

## Screenshots/Videos

N/A

- [x] All checks have passed

### DIFF
--- a/elements/timecontrol/src/main.ts
+++ b/elements/timecontrol/src/main.ts
@@ -113,6 +113,9 @@ export class EOxTimeControl extends LitElement {
     this._updateStep(0);
   }
 
+  /**
+   * Gets or sets the current selected time as a string.
+   */
   get currentTime() {
     return this.animationValues[this._newTimeIndex];
   }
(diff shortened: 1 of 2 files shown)


silvester-pari (changes requested):
Thanks for working on this! I guess it would be more web component-y to have a reactive property instead of a method. Something like
```
eoxTimeControl.currentTime = <timestring>
```
This would be solved via a getter/setter.
This new property would be also updated, so the current state can be read with `eoxTimeControl.currentTime`.

silvester-pari (approved):

PR #621, opened 2024-02-06
feat: initial layout element setup

silvester-pari:
## Implemented changes

This is the initial release of the new `eox-layout` element. It should facilitate the creation of user interface layouts based on a 12/12 grid. Initial features:
- `eox-layout` and `eox-layout-item` elements
- `gap` attribute for `eox-layout`
- configurable for `eox-layout-item`: **x** (coordinate), **y** (coordinate), **w**(idth) and **h**(eight)
- it has basic reactivity, meaning once attributes change, the layout re-renders

Future evolution might include features such as creating a layout from JSON, responsive layouts etc.

## Screenshots

screencapture-localhost-6006-2024-02-08-12_15_08

- [x] All checks have passed

### DIFF
--- a/elements/layout/stories/grid.js
+++ b/elements/layout/stories/grid.js
@@ -2,6 +2,7 @@ import { html } from "lit";
 import "../src/main";
 import { STORIES_LAYOUT_STYLE } from "../src/enums";
 
+// Generate one [x,y] array for each slot in a 12x12 grid
 const renderItems = () => {
   let items = [];
   for (let x = 0; x < 12; x++) {
--- a/elements/layout/stories/grid.js
+++ b/elements/layout/stories/grid.js
@@ -1,0 +1,30 @@
+import { html } from "lit";
+import "../src/main";
+import { STORIES_LAYOUT_STYLE } from "../src/enums";
+
+const renderItems = () => {
+  let items = [];
+  for (let x = 0; x < 12; x++) {
+    for (let y = 0; y < 12; y++) {
+      items.push([x, y]);
+    }
+  }
+  return items;
+};
+
+export const Grid = {
+  args: {},
+  render: () => html`
+    <!-- Render eox-layout component -->
+    <eox-layout style="${STORIES_LAYOUT_STYLE}">
+      ${renderItems().map(
+        ([x, y]) =>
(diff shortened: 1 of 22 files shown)


A-Behairi (approved):

lubojr (approved):

srijitcoder (approved):
Wouldn't it be great to include a top-level comment in the files located in `elements/layout/stories/*.js` (* = grid/primary/gap)? 

This approach could significantly clarifies the purpose and expected output of each file, particularly in light of the existing `x,y,w,h` loop inside `renderItem` function that might be confusing at beginning.

StefanBrand:
I'm wondering about how content with a fixed aspect ratio would look like, i.e. a square satellite image. Will it be able to have grid cells that are completely square (even if this means that the 12 rows would overflow the viewport)?

silvester-pari:
> I'm wondering about how content with a fixed aspect ratio would look like, i.e. a square satellite image. Will it be able to have grid cells that are completely square (even if this means that the 12 rows would overflow the viewport)?

@StefanBrand good question, and a potential first feature expansion. Currently the grid is contained within the parent, which means a square grid item would squish all other items a bit, and the items of the entire row would be square (applied to item 0/4 in this example):

image

You can also specify the aspect ratio of a single item, but that would cause some overflow issues with other items.

But we could consider letting the grid overflow in case there is a square item.

silvester-pari:
Thanks everyone for the feedback!

elements/layout/stories/grid.js
```diff
@@ -0,0 +1,31 @@
...
+
+// Generate one [x,y] array for each slot in a 12x12 grid
+const renderItems = () => {
+  let items = [];
+  for (let x = 0; x < 12; x++) {
+    for (let y = 0; y < 12; y++) {
+      items.push([x, y]);
+    }
+  }
+  return items;
```
StefanBrand, line 13: I found a cool algorithm for this:

```js
const gridNumbers = [...Array(11).keys()].map(n => n+1);
return gridNumbers.flatMap(i => gridNumbers.map(j => [i, j])) 
// [[1,1],[1,2],[1,3],[1,4],[1,5],[1,6],[1,7],[1,8],[1,9],[1,10],[1,11],[1,12],[2,1],...]
```

PR #631, opened 2024-02-08
feat: support non-tiled sources

### LINKED ISSUE #629: Support ImageWMS/XYZ layers
Currently the timecontrol only supports tiled layer source: https://github.com/EOX-A/EOxElements/blob/be45bf2eaf22235f35a4716efee03d8abff897dd/elements/timecontrol/src/main.ts#L208

For other source types like ImageWMS etc. this won't work as those don't have any `tileUrlFunction`

silvester-pari:
## Implemented changes
Edit: after some discussion, we opted to just support param updates instead of maintaingin both tile-specific cases and other cases. This also removes the usage of OL-internal API (which we should do anyway). This setup is now much simpler, and for special/complicated cases we would use a STAC-based approach (e.g. each "time" entry is a separate STAC item with a self-contained visualization url).

~This adds support for non-tiled sources by checking if a `tileUrlFunction` exists. If not, it is assumed to be a non-tiled source, and the params are used instead. The reason behind this is that ImageWMS sources don't have a url loader function and generally use urls + params instead.~

~We could use url + params also for tiled sources, but with the current setup, we would allow deeper level control of url manipulation in tiled source urls. Although I am open to replace all by the param updating via `updateParams`.~

~More details: Both ImageWMS and TileWMS have a [updateParams](https://openlayers.org/en/latest/apidoc/module-ol_source_TileWMS-TileWMS.html#updateParams) function, which allows setting new params like TIME, LAYER etc. This can be used to switch between dates that an endpoint provides. But, additionally, a TileWMS source also offers a [setTileUrlFunction](https://openlayers.org/en/latest/apidoc/module-ol_source_TileWMS-TileWMS.html#setTileUrlFunction) function, which (if I understand correctly), allows a deeper level of url manipulation, not limited to params. If we have use cases where this is needed we can keep both. If we say we only need to upgrade params, then both tiled and non-tiled sources could use `updateParams()`. I am not an expert on this, so I am unsure.~

Fixes #629

- [x] All checks have passed

### DIFF
--- a/elements/timecontrol/src/main.ts
+++ b/elements/timecontrol/src/main.ts
@@ -47,13 +47,13 @@ export class EOxTimeControl extends LitElement {
   slider: boolean;
 
   /**
-   * Original params (in case of a image source).
+   * Original params of layer source
    */
   @property()
   private _originalParams: object;
 
   /**
-   * Hides the play button if set.
+   * Hides the play button if set
    */
   @property({ attribute: "disable-play", type: Boolean })
   disablePlay: boolean;
--- a/elements/timecontrol/src/main.ts
+++ b/elements/timecontrol/src/main.ts
@@ -63,11 +63,17 @@ export class EOxTimeControl extends LitElement {
   urlFunction: (url: string) => string;
 
   /**
-   * Original tile url function of the source.
+   * Original tile url function of the source (in case of a tiled source).
    * Used to get the correct TileGrid-Values, while manipulating certain parts of the URL.
    */
   @property()
   private _originalTileUrlFunction: UrlFunction;
+
+  /**
+   * Original params (in case of a image source).
+   */
+  @property()
+  private _originalParams: object;
 
   /**
    * Hides the play button if set.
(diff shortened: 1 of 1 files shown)


lubojr (approved):
We should allows a deeper level of url manipulation. I don't think that just updating params is going to be enough for all use cases for wierd remote server implementations that we might encounter. Better to be safe if the OpenLayers have the functionality.

elements/timecontrol/src/main.ts
```diff
@@ -161,33 +167,42 @@ export class EOxTimeControl extends LitElement {
...
+            src.substring(src.indexOf("?"))
+          );
+          searchParams.set(
+            this.animationProperty,
+            this.animationValues[this._newTimeIndex]
+          );
+
+          return (
+            src.substring(0, src.indexOf("?") + 1) + searchParams.toString()
+          );
```
StefanBrand, line 193: construct a [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL) instead of manually parsing `src`. It allows to manipulate its `searchParams` directly.

StefanBrand (changes requested):
In general, I vouch for keeping things simple. Unless there is a specific usecase for `tileUrlFunction`, it should suffice to manipulate params.

silvester-pari:
> We should allows a deeper level of url manipulation. I don't think that just updating params is going to be enough for all use cases for wierd remote server implementations that we might encounter. Better to be safe if the OpenLayers have the functionality.

@lubojr The issue I found was that this is only available for tiled sources, not for image sources. Unless I am mistaken. This is why currently tileUrlFunction for tiled sources, and params manipulation for non-tiled

> In general, I vouch for keeping things simple. Unless there is a specific usecase for tileUrlFunction, it should suffice to manipulate params.

@StefanBrand I think we had a usecase in eodash for this, but in future eodash will work differently, since each "date" is ideally a self-contained STAC item, so the functionality implemented here in timecontrol would become obsolete for that.

I'll try to make param manipulation the default but also optionally allow deeper url manipulation, and split things up better in the code.

PR #668, opened 2024-02-20
Feat: fire custom event when time is changed in timeControl

viktoraronfarkas:

- [ ] All checks have passed

### DIFF
--- a/elements/timecontrol/src/main.ts
+++ b/elements/timecontrol/src/main.ts
@@ -150,6 +150,13 @@ export class EOxTimeControl extends LitElement {
       [this.animationProperty]: this.animationValues[this._newTimeIndex],
     });
     this.requestUpdate();
+
+    let event = new CustomEvent('onTimeChange', {
+      detail: {
+        time: this.animationValues[this._newTimeIndex],
+      }
+    });
+    this.dispatchEvent(event);
   }
 
   /**


elements/timecontrol/src/main.ts
```diff
@@ -150,6 +150,18 @@ export class EOxTimeControl extends LitElement {
...
     this.requestUpdate();
+
+    /**
+     * Triggers when *currentTime* is updated.
+     * `event.detail.time` returns the new *currentTime* value.
+     */
+    this.dispatchEvent(
+      new CustomEvent("onTimeChange", {
+        detail: {
+          time: this.animationValues[this._newTimeIndex],
```
silvester-pari, line 161: If we only pass the "time" here we could directly set it as the `detail`. On the other hand, we could also pass other things such as the index

silvester-pari: Another thought, as the property is called `currentTime`, it might be better to call it this way? So you could probably also do 
```
detail: {
  currentTime: this.currentTime
}
```
?

silvester-pari (changes requested):

elements/timecontrol/src/main.ts
viktoraronfarkas: I think using `detail: {
  currentTime: this.currentTime
}` would be the most intuitive. We could pass the index, but on the other hand it is really easy to get the index from `currentTime`, using `Array.indexOf()`.

silvester-pari (approved):
Thanks for your first contribution :-) Looks fine, good addition!

PR #690, opened 2024-02-26
Feat: Add markdown rendering via `markdown-it` as string and URL

srijitcoder:
## Implemented changes
- Basic markdown render using 2 attributes - 
  - `markdown`: Add markdown as string  
  - `markdown-url`: Add markdown as URL 
- Fetch `markdown` through API and render it into dom.
- Usage:

```html
<eox-storytelling markdown="[...md...]"></eox-storytelling>
// or
<script>
  document.querySelector("eox-storytelling").markdown = "[...md...]";
</script>
```

or 

```html
<eox-storytelling markdown-url="//path/to/url"></eox-storytelling>
// or
<script>
  document.querySelector("eox-storytelling").markdownURL = "//path/to/url";
</script>
```

## Screenshots/Videos

image

- [ ] ~I have added a test related to this feature/fix~
- [x] All checks have passed


.storybook/main.js
```diff
@@ -23,5 +23,6 @@ const config = {
     autodocs: "tag",
     toc: true,
   },
+  staticDirs: ["../public"],
```
silvester-pari, line 26: Is there any way of having this static folder within the `/elements/storytelling/stories/` folder? It would make more sense having it there. If not, maybe we can introduce a structure like `/public/stories/storytelling/sample.md`.

elements/storytelling/src/main.js
```diff
@@ -1,12 +1,66 @@
 import { LitElement, html } from "lit";
+import markdownit from "markdown-it";
+import { unsafeHTML } from "lit/directives/unsafe-html.js";
+import { loadMarkdownURL } from "./helpers/index";
+import mainStyle from "../../../utils/styles/dist/main.style";
+const md = markdownit({ html: true });
```
silvester-pari, line 6: If we enable `html` from the start, then it would probably no harm in also adding some sanitation. E.g. using [DOMPurify](https://github.com/cure53/DOMPurify) - see https://github.com/markdown-it/markdown-it/blob/master/docs/security.md.
Or, we don't enable `html` but handle each different exception via plugins.

silvester-pari (changes requested):

.storybook/main.js
srijitcoder: Yeah we can rather than passing `["../public"]` we can pass `["../elements/storytelling/stories/public/"]`

elements/storytelling/src/main.js
srijitcoder: Sure I thought I will do it when we start implementing things with actual markdown patterns. No issue I will add it right now.

silvester-pari (approved):

srijitcoder (commented):
Looks good to me 👍🏼

PR #698, opened 2024-02-27
fix(docs): improve args/props rendering

silvester-pari:
## Implemented changes

When generating the autodocs code, Storybook scraps properties added via the lit `.<property>` syntax. The previous solution worked for some cases, but it
- wrongly applied args to the elements (the story element would get all args) and it
- applied args that weren't even used

The new approach goes through the raw lines line by line and tries to figure out which arg belongs to which element.

### Before
image

### After
image

### Before
image

### After
image

- [x] All checks have passed

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -37,21 +37,51 @@ const preview = {
       page: DocumentationTemplate,
       source: {
         transform: (code, storyContext) => {
+          /**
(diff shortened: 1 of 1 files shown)


viktoraronfarkas (commented):
looks good to me, thanks! will try it out in the agri implementation soon💪

PR #704, opened 2024-03-03
feat: add dereference functionality to eox-jsonform

radupasparuga:
## Implemented changes
Added [json-schema-ref-parser](https://github.com/APIDevTools/json-schema-ref-parser) to support dereferencing in `eox-jsonform`

- [ ] All checks have passed

### DIFF
--- a/elements/jsonform/jsonform.stories.js
+++ b/elements/jsonform/jsonform.stories.js
@@ -11,6 +11,11 @@ const externalSchema =
 const externalStartVals =
   "https://raw.githubusercontent.com/EOX-A/EOxElements/main/elements/jsonform/examples/catalogStartVals.json";
 
+const dereferenceSchema =
+  "https://raw.githubusercontent.com/EOEPCA/open-science-catalog-metadata-testing/ui-schemas/schemas/variables/children.json";
+const dereferenceStartVals =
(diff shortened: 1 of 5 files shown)


silvester-pari (changes requested):
Before merging, we need to clean up a bit:
- use jsonform as (built) package import in  layercontrol instead of directly importing it (should be done in a different PR)
- try to add the polyfills in as few places as possible, but still tackling all necessary locations (storybook server, cypress component server, built elements for e2e tests)

silvester-pari:
This turns out to be more tricky than anticipated. It works with the current setup, but it is quite ivasive as we have to define the polyfills/shims in 3 different locations:
- In the general vite config for Storybook dev server
- In the component test support file for component tests
- In the element vite config for e2e tests and production bundling

I tried having everything contained in the element itself, but didn't manage to get it working.

Maybe, in the end, we need to dereference the schema server-side...

Let's keep this open a bit longer and maybe someone else has an idea.

silvester-pari:
Maybe the dereferencing can be done in a GitHub action instead?

santilland:
Seeing that we quite often have browser support issues we have to take care of, i would really vote in favor of not including these dependencies into the package, as anywhere we include it we would need to be very vary of browser support.
As far as i understand it is possible to "statically" dereference the schemas o have them in a ready to read form. I would propose to try to do this in a build step when some schema is committed to the repository. This would also solve potential external references not being available temporarily.

silvester-pari:
Closing for now, since the better approach would probably be to add dereferencing into a GitHub action.

Issue #702, opened 2024-02-29
Support multiple projections


silvester-pari:
Add functionality to init the map with a different projection (like `EPSG:4326`). Bonus points if we implement a projection switcher that allows this on the fly.

lubojr:
Will be very much needed for our planned polar dashboard v5 which has datasets which are both interesting in some Polar projection and EPSG:4326 so on-demand-switcher (replacing existing used OL view with a one with a different projection).

RobertOrthofer:
I will look into it as soon as possible

silvester-pari:
Implemented in https://github.com/EOX-A/EOxElements/pull/730

PR #730, opened 2024-03-07
feat: switch projections & allow advanced projections

RobertOrthofer:
This PR brings the usage of projections, as described in #702.

Out-of-the-box, `EPSG:4326` is supported in addition to the standard `EPSG:3857`. Other projections can be used by calling the helper-function `registerProjection` first, including a name and proj4 definition string. These can be obtained by via services like [epsg.io](epsg.io), this must be handled by the parent application.

Switching the projection is possible by setting the `projection`-attribute. A new `view` is created when doing so, listener functions are moved over to the new view. It is important that set listeners retrieve the view via `evt.target`, instead of `this`.

- [x] All checks have passed

### DIFF
--- a/elements/map/helpers.ts
+++ b/elements/map/helpers.ts
@@ -11,6 +11,8 @@ import { getUid } from "ol";
 import { DrawEvent } from "ol/interaction/Draw";
 import { DragAndDropEvent } from "ol/interaction/DragAndDrop";
 import Feature from "ol/Feature";
+import proj4 from "proj4";
+import { register } from "ol/proj/proj4";
 
 /**
  * Specifies the options for reading features with defined source and target projections.
@@ -212,4 +214,12 @@ export function isTopoJSON(text: string): boolean {
   } catch (e) {
     return false;
   }
+}
+
+export function registerProjection(
+  name: string,
+  projection: string | proj4.ProjectionDefinition
+) {
+  proj4.defs(name, projection);
+  register(proj4);
 }
--- a/elements/map/helpers.ts
+++ b/elements/map/helpers.ts
@@ -12,7 +12,7 @@ import { DrawEvent } from "ol/interaction/Draw";
 import { DragAndDropEvent } from "ol/interaction/DragAndDrop";
 import Feature from "ol/Feature";
 import proj4 from "proj4";
-import { register } from "ol/proj/proj4";
+import { fromEPSGCode, register } from "ol/proj/proj4";
 
 /**
  * Specifies the options for reading features with defined source and target projections.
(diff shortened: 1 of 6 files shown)


elements/map/helpers.ts
```diff
@@ -213,3 +215,11 @@ export function isTopoJSON(text: string): boolean {
     return false;
   }
 }
+
+export function registerProjection(
+  name: string,
+  projection: string | proj4.ProjectionDefinition
```
silvester-pari, line 221: Wouldn't it be a very nice convenience function to make this async and fetch the definition for the user from https://epsg.io/?format=json&q=? Like in https://openlayers.org/en/latest/examples/reprojection-by-code.html.

elements/map/main.ts
```diff
@@ -208,6 +212,57 @@ export class EOxMap extends LitElement {
...
+  }
+
+  /**
+   * projection of the map view as SRS-identifier (e.g. EPSG:4326)
+   */
+  @property({ attribute: "projection", type: String })
+  set projection(projection: ProjectionLike) {
+    const oldView = this.map.getView();
+    if (projection && projection !== oldView.getProjection().getCode()) {
+      const newCenter = transform(
```
silvester-pari, line 231: I didn't manage to successfully test this. When changing the projection from e.g. 3857 to 4326, then the zoom is reset to initial. I tried in storybook:

```
      <button
        @click=${() => {
          const eoxMap = document.querySelector("eox-map");
          eoxMap.setAttribute("projection", "EPSG:4326");
        }}
      >
        change
      </button>
```

Initial center: `center: [16.8, 48.2],`

silvester-pari (changes requested):
Looks good in general, thank you! Please see the two comments, and please add a story showcasing this new functionality :slightly_smiling_face:

silvester-pari (approved):

elements/map/helpers.ts
RobertOrthofer: i added this as another helper functionality, utilizing `fromEPSGCode`

elements/map/main.ts
RobertOrthofer: i couldn't reproduce this behaviour with zoom, but i did find weird interactions with the initial center, which came down with weird interactions and timings when setting the center of the view and setting the target of the map... this was quite tough to find, i restructured the `center` to be in line with the other attributes (with a getter and setter instead of watching for changes in the `updated`-hook), and setting the center once after the map was added to the `DOM`.

Also added an entry to the story book

lubojr:
@RobertOrthofer Thank you very much for this new feature! :rocket: 
I have an implementation detail question - what will happen now if the old "center" of current view can not be projected into the new projection? Will it not throw an openlayers exception?

Example: When using a polar stereographic projection centered on pole and decide to switch my map to EPSG:3857 which has projection bounds geographically around 85 degrees so my original center will be unprojectable. Would it be a good idea to in that case set center to new projection center ?

RobertOrthofer:
that is a great question, do you have an EPSG-code to test this out? I have intentionally left out setting the `extent` of the view to the `extent` of the projection for now, as this might limit some very edge cases

lubojr:
@RobertOrthofer Let's try this out. 

For example the North Polar stereographic can be used: 

proj4 def:
```
{
  name: 'EPSG:3411',
  def: '+proj=stere +lat_0=90 +lat_ts=70 +lon_0=-45 +x_0=0 +y_0=0 +a=6378273 +b=6356889.449 +units=m +no_defs +type=crs',
  extent: [-3314763.31, -3314763.31, 3314763.31, 3314763.31],
}
```

RobertOrthofer:
I tried this with the `registerProjectionFromCode`-helper and it looks like it always manages to reproject the center. Obviously, if the center is way off the projections bounds, the map becomes quite distorted, but usable...
I think for now this would be the responsibility of the application to move the map into the area of interest beforehand, as limiting the extent might not always be the desired behaviour

lubojr:
Perfect! Thanks for checking!

lubojr (approved):

silvester-pari (approved):

PR #737, opened 2024-03-11
feat: reactive props

### LINKED ISSUE #662: Allow setting the schema / start vals after initialization
Add a native custom element getter/setter instead of `setSchema`/ `setDefaultValues`, and update/re-render the form when this properties are changed (currently the form is only rendered on `firstUpdated`.

Also, add a test showcasing that the schema/vals can be changed at a later time.

radupasparuga:
Fixes #662 
## Implemented changes
- Replace `firstUpdated` with `updated` (https://github.com/EOX-A/EOxElements/pull/737/commits/20baaf7beeb81226a826d8b44cdf49d46c0ae2b2) to allow `schema` and `startVals` to be changed after the first render
- user proper setters and getters for `schema` and `startVals` (https://github.com/EOX-A/EOxElements/pull/737/commits/a8ccd5fe3826eae4a68a2327659693628d3f9a19)
- add test to check if switching `schema` and `startVals` works after first render (https://github.com/EOX-A/EOxElements/pull/737/commits/a3caa4e720b6e25ad744fcf693f9ef2a0ad1caee and https://github.com/EOX-A/EOxElements/pull/737/commits/aab4821947b297908fb27f92bb4ba2cfe0fff594)

- [x] All checks have passed


elements/jsonform/src/main.js
```diff
@@ -88,34 +88,31 @@ export class EOxJSONForm extends LitElement {
   /**
    * Get The JSON schema for rendering the form
```
silvester-pari, line 89: Could you update this description? Since now it does not only get, but also set. In the Storybook Docs it could be a bit confusing otherwise:

image

elements/jsonform/src/main.js
```diff
@@ -88,34 +88,31 @@ export class EOxJSONForm extends LitElement {
...
-  setSchema(newSchema) {
-    this.schema = newSchema;
+  set schema(newSchema) {
+    let oldValue = this._schema;
+    this._schema = newSchema;
+    this.requestUpdate("schema", oldValue);
   }
 
   /**
    * Get default value for rendering the form
```
silvester-pari, line 104: Same as above

elements/jsonform/src/main.js
```diff
@@ -146,9 +143,9 @@ export class EOxJSONForm extends LitElement {
     });
   }
 
-  async firstUpdated() {
-    this.setSchema(await this.parseProperty(this.schema));
-    this.setDefaultValues(await this.parseProperty(this.startVals));
+  async updated() {
+    this.schema = await this.parseProperty(this.schema);
```
silvester-pari, line 147: If I understand correctly, the `updated()` function triggers every time the component is updated and/or re-rendered. Maybe it would make sense to check if the schema is already set? Something like if the current schema is already the same as the property, do nothing. Same for startVals.

silvester-pari (changes requested):

elements/jsonform/jsonform.stories.js
```diff
@@ -20,7 +20,7 @@ export default {
   render: (args) => html`
     <eox-jsonform
       .schema=${args.schema}
-      .startVals=${args.startVals}
+      .value=${args.startVals}
```
silvester-pari, line 23: Let's please also rename this in the storybook args, so `args.value` instead of `args.startVals`

silvester-pari (changes requested):

elements/jsonform/jsonform.stories.js
radupasparuga: Resolved in https://github.com/EOX-A/EOxElements/pull/737/commits/f6d29b84aa3a6c6cc3d35b2c7cfbc76c0bd15970

silvester-pari (approved):

Issue #562, opened 2024-01-09
Geolocation with feature display


silvester-pari:
Consider offering a convenient way to access and display geolocation on the map - https://openlayers.org/en/latest/examples/geolocation.html

StefanBrand:
User requirement was stated as follows:

>>>
Geoposition / Location anzeigen und anwenden

anwenden = automatically zoom to position on startup in ParcelMap, e.g. during field visits on tablet
>>>

silvester-pari:
@StefanBrand could this be solved with a map control? So the user clicks on a button within the map and is zoomed to the position?

StefanBrand:
Since we must support both mobile and in-house usage, we should start off with a button that users can push on demand.

PR #769, opened 2024-03-20
feat: geolocation control

RobertOrthofer:
## Implemented changes

This PR brings a custom Geolocation-control as defined in #562, that can be used just like any other `ol`-control. The user has to agree to the location-prompt by the browser, which will only show up when served via https (which is also true for testing, workarounds like ngrok or mkcert will be needed).

The `centerWhenReady`-prop will pan the view to the user position on the first update of the geolocator.

Rotation is not implemented by default, but the `heading`-property of the position feature is set, so by setting the `style`-property of the geolocation control a rotation get be set by accessing that property via style expressions.
(to enable tracking the heading, `highAccuracy` and `trackHeading` have to be set).

With `trackAccuracy` the accuracy feature will be shown in addition to the position feature, indicating possible low accuracy of the sensor.

A custom Icon for the button can be set via the `buttonIcon`-prop. The button element get be retrieved via `Geolocation.getElement()`

- [x] All checks have passed

### DIFF
--- a/elements/map/main.ts
+++ b/elements/map/main.ts
@@ -14,7 +14,7 @@ import {
   addOrUpdateControl,
   controlDictionary,
   controlType,
-} from "./src/controls";
+} from "./src/controls/controls";
 import { buffer } from "ol/extent";
 import "./src/compare";
 import {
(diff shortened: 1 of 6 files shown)


RobertOrthofer:
There might be a cleaner way for the positioning (css injection?) of the button element. Right now it is assumed that the default zoom-buttons are available, as this is most likely the most common setup.

silvester-pari (approved):
Looks great!

StefanBrand (approved):

PR #791, opened 2024-03-27
Feat: use eox-jsonform instead of monaco editor

srijitcoder:
## Implemented changes

- Used `eox-jsonform` to implement `mardown-editor` using `simple-mde`
- Replaced `monaco editor` completely 
- Added two custom elements 
  - Insert Simple Map 
  - Insert Map Tour

## Screenshots/Videos

image

- [x] All checks have passed

### DIFF
--- a/elements/storytelling/src/components/editor.js
+++ b/elements/storytelling/src/components/editor.js
@@ -1,5 +1,7 @@
 import { LitElement, html } from "lit";
-import initEditor from "../helpers/editor";
+import "../../../jsonform/src/main.js";
+import "https://cdn.jsdelivr.net/simplemde/latest/simplemde.min.js";
+// TODO import helpers for moving editor etc.
 
 // Define LitElement for the editor
 class StoryTellingEditor extends LitElement {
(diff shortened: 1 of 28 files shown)


elements/storytelling/src/components/editor.js
```diff
@@ -1,29 +1,22 @@
 import { LitElement, html } from "lit";
-import initEditor from "../helpers/editor";
+import "../../../jsonform/src/main.js";
```
silvester-pari, line 2: We shouldn't import eox-jsonform as this would increase bundle size. We should instead inform the developer that if the editor is used, also eox-jsonform needs to be installed.
See https://github.com/EOX-A/EOxElements/blob/main/elements/layercontrol/src/components/layer-config.js#L113-L115 for an example how this is handled in eox-layercontrol.

silvester-pari (changes requested):

elements/storytelling/src/components/editor.js
srijitcoder: Ahh, I didn't know we were informing the developer about the import of other `eox-elements`. Thanks for informing me, will look into this in future as well.

srijitcoder: Resolved ✅

silvester-pari (approved):

PR #878, opened 2024-04-22
chore(test): Storybook test integration

radupasparuga:
## Implemented changes
This PR introduces a new test format (additionally to component and e2e tests) to the PR check pipeline.
This new test loops through all Storybook stories and fails in case of console errors.

Screenshot 2024-05-16 at 15 42 14

With the new action we managed to detect the above error in the Stacinfo custom slot content story and fix it!

Screenshot 2024-05-16 at 15 54 51

After the fix we can see that all 67 tests are passing.

Average time to run through every story is ~3 minutes but that won't happen on every PR, in [cypress/e2e/storybook.cy.js](https://github.com/EOX-A/EOxElements/pull/878/files#diff-5e50c1d03b3df3173950c3b11ae292640c8dc7e4d92e4de7529f6d55ddb42d91) we check the git diff (passed as an environment variable) so that we only test stories of elements that were changed in the current branch with the exception that if the `cypress` folder has been changed it will run through every story (heavy inspiration from [cypress.config.ts](https://github.com/EOX-A/EOxElements/blob/main/cypress.config.ts#L5-L22)).

- [x] All checks have passed


silvester-pari (approved):
Thank you, this is great! It adds another layer of tests, and it did indeed find some errors in Storybook stories (some of which have been fixed in the meantime). Since we have to write the stories anyways, it makes sense to also use them as tests.

lubojr (approved):
This is really helpful enhancement to the storybook setup and it could catch future non-working stories.

Just a miscellaneous question - is the `excludeSpecPattern='[\"**/layercontrol.cy.js\"]' necessary?`

radupasparuga:
> This is really helpful enhancement to the storybook setup and it could catch future non-working stories.
> 
> Just a miscellaneous question - is the `excludeSpecPattern='[\"**/layercontrol.cy.js\"]'` necessary?

In [cypress.config.ts](https://github.com/EOX-A/EOxElements/pull/878/files#diff-3f62e885350a48939f240b9daf40f6b06d41da7751f2b613238fc7669ac0a1b2R31) we exclude the storybook spec so it doesn't run at the same time with the E2E tests (storybook tests take 2-3 minutes so it makes life easier when trying to only run E2E tests) but in order to actually run the spec we need to overwrite the config via the CLI and in that case I'm excluding a fictitious `layercontrol` spec.
Not the cleanest solution but it gets the job don 😅

PR #903, opened 2024-04-30
Fix: Editor height alignment \

srijitcoder:

- [ ] All checks have passed

### DIFF
--- a/elements/storytelling/src/components/editor.js
+++ b/elements/storytelling/src/components/editor.js
@@ -1,5 +1,5 @@
 import { LitElement, html } from "lit";
-import { initEditorEvents } from "../helpers";
+import { initEditorEvents, positionEditor } from "../helpers";
 import { EDITOR_SCHEMA } from "../enums";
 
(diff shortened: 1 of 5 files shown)


santilland:
Yea, a bit more involved then expected...
I guess this is the way to go, not much we can do if we want to be able to drag it around and be "floating" on top of the actual story.
@srijitcoder you changed to to be fixed against the bottom instead of top in the style?
In any case, thanks!

srijitcoder:
> Yea, a bit more involved then expected... I guess this is the way to go, not much we can do if we want to be able to drag it around and be "floating" on top of the actual story. @srijitcoder you changed to to be fixed against the bottom instead of top in the style? In any case, thanks for the solution!

Yes, this will work. I've positioned the editor at the bottom and initialized its height upon loading, based on the parent element's position and other miscellaneous elements like padding and `story-telling` navigation height if it's enabled.

silvester-pari (approved):

Issue #835, opened 2024-04-09
Selecting an item in ItemFilter when two items with same title does not show selected state correctly


lubojr:
If I select an item in ItemFilter and there is another item with the same title (in a different group for example), the wrong radio button can be randomly "selected".

https://github.com/EOX-A/EOxElements/assets/19269667/dcd551a2-3b53-4f7d-81f0-dcd5f3ebd9b4

PR #910, opened 2024-05-02
fix: improve incorrect selection highlighting

spectrachrome:
## Implemented changes
As described by Lubo in #835, with certain indicators, the selection in the Item Filter renders incorrect additional highlightings at other positions than the active radio button.

This is because the comparison responsible for the highlighting only compared the titles of the items, resulting in additional highlightings for any others that had the same title. This PR changes the comparison to use the item IDs directly, which is definitely an improvement, however this does still result in collisions if an indicator happens to be present in more than a single theme – highlighting the same indicator in multiple theme dropdowns.

For Lubo, that behavior is fine, but I am a bit uncomfortable about the fact that our code generates two objects with identical ID fields. Is this possibly something to investigate, and to implement a fix for that enables properly unique IDs?

## Screenshots/Videos

https://github.com/EOX-A/EOxElements/assets/94269527/8a1551a6-a1fc-4b04-a3e0-69c63e59c8e6

- [x] All checks have passed

### DIFF
--- a/elements/itemfilter/src/main.ts
+++ b/elements/itemfilter/src/main.ts
@@ -344,9 +344,7 @@ export class EOxItemFilter extends TemplateElement {
                   class="result-radio"
                   name="result"
                   id="${<string>item.id}"
-                  ?checked=${this.selectedResult?.[
-                    this._config.titleProperty
-                  ] === item[this._config.titleProperty] || (nothing as null)}
+                  ?checked=${this.selectedResult?.id === item.id || (nothing as null)}
                   @click=${() => {
                     this.selectedResult = item;
                     this._config.onSelect(item);
--- a/elements/itemfilter/src/main.ts
+++ b/elements/itemfilter/src/main.ts
@@ -344,7 +344,8 @@ export class EOxItemFilter extends TemplateElement {
                   class="result-radio"
                   name="result"
                   id="${<string>item.id}"
-                  ?checked=${this.selectedResult?.id === item.id || (nothing as null)}
+                  ?checked=${this.selectedResult?.id === item.id ||
+                  (nothing as null)}
                   @click=${() => {
                     this.selectedResult = item;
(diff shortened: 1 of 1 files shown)


lubojr (approved):
Thank you very much for a quick identification of the problematic area in the code and providing a fix. Without too much context, I am very much fine with this solution and am up for a merge.

@silvester-pari Should we continue with further into the uniqueness of ID topic?

silvester-pari:
Thanks for looking into this, however currently there is no guarantee that the items will have an `id`. If we want to make it analogous to the title, we'd need to add an `idProperty`, so the developer can choose one of the item's fields to act as an unique id.
Another possible strategy would be to use the initial index as an identifier; we would get into trouble if the items change (items are pushed in/out), but this could be mitigated with a "hard" reset (e.g. if items are changed and the `apply()` function is called, deselect any previously selected items, then create new ids based on index). This strategy would remove the need for a unique property of the items.
We could also do a mix, the second strategy as a fallback for the first - I am open for discussion :-)

In any case, the current solution is not ideal as it hardcodes a potential `id` property, and I'd like to solve that first (besides the tests failing).

silvester-pari (changes requested):
Thanks for looking into this, however currently there is no guarantee that the items will have an id. If we want to make it analogous to the title, we'd need to add an idProperty, so the developer can choose one of the item's fields to act as an unique id.
Another possible strategy would be to use the initial index as an identifier; we would get into trouble if the items change (items are pushed in/out), but this could be mitigated with a "hard" reset (e.g. if items are changed and the apply() function is called, deselect any previously selected items, then create new ids based on index). This strategy would remove the need for a unique property of the items.
We could also do a mix, the second strategy as a fallback for the first - I am open for discussion :-)

In any case, the current solution is not ideal as it hardcodes a potential id property, and I'd like to solve that first (besides the tests failing).

PR #912, opened 2024-05-03
feat: create a new `EOxGeoSearch` element for gazetteer search

spectrachrome:
## Implemented changes

This pull request contains a new `EOxGeoSearch` element which provides an autocompleted search input for geographic locations. The element comes with a configurable button mode which can be enabled with the boolean attributes `button` and, additionally, `small`, to use the button in a map toolbar.

Real-time OpenCage API access is enabled by setting both the `endpoint` and `key` attributes. Optionally, a `limit` attribute may be set to restrict the number of fetched locations.

Furthermore, the alignment of the input and search results can be configured to align straight or in a 90-degree angle in any direction with the `input-direction` and `results-direction` attributes and the following directions:

- `left`
- `up`
- `right`
- `down`

## Screenshots/Videos

Screen Recording 2024-05-21 at 12 11 02

- [x] For new elements: I have created a story showcasing this new element
- [x] All checks have passed


elements/geosearch/main.js
```diff
@@ -0,0 +1,375 @@
...
+       * - `right`
+       * - `bottom`
+       *
+       */
+      resultsDirection: { type: String, default: "rightx" },
+      /**
+       * The direction of the search results relative to the input field, with the same options as described above.
+       *
+       */
+      secondaryDirection: { type: String },
```
silvester-pari, line 105: This doesn't seem to be used

elements/geosearch/main.js
```diff
@@ -0,0 +1,375 @@
...
+      /**
+       * The direction of the results box relative to the input, with the following options:
+       *
+       * - `left`
+       * - `top`
+       * - `right`
+       * - `bottom`
+       *
+       */
+      resultsDirection: { type: String, default: "rightx" },
```
silvester-pari, line 100: Should probably be `right`, no t`rightx`

elements/geosearch/main.js
```diff
@@ -0,0 +1,375 @@
...
+import styles from "./style";
+
+class EOxGeoSearch extends LitElement {
+  static get properties() {
+    return {
+      /**
+       * Internal storage of OpenCage API data after a successful API request.
+       *
+       */
+      _data: { attribute: false },
```
silvester-pari, line 17: Private properties should not appear in the API docs (try using `@` instead of `_` for natively private properties)

image

elements/geosearch/main.js
silvester-pari, line 100: In general, attributes can't have camelCase - needs e.g. `attribute: "results-direction"`

elements/geosearch/main.js
silvester-pari, line 105: `attribute: "secondary-direction"`

silvester-pari, line 105: maybe calling it "listDirection/list-direction" or "resultDirection/result-direction" would be easier to understand? :thinking:

elements/geosearch/main.js
```diff
@@ -0,0 +1,375 @@
...
+            align-items: ${this.getVerticalAlign()};
+        "
+      >
+        <button
+          class="${this.button ? "" : "hidden"}"
+          style="
+                        margin-${this.getMarginDirection(this.direction)}: 12px;
+                        background: ${this._isInputVisible
+            ? "#0078CE"
+            : "#004170"};
```
silvester-pari, line 262: It would be necessary to allow the parent application to override the styling via CSS variables, here the color seems to be pretty hardcoded.

silvester-pari (changes requested):
Hi, thanks for this addition. Some general remarks:
- please don't use/reference mock data inside `main.js`, mocking should only be done for tests/stories and not inside the main JS file
- if needed, consider using a simpler template url string (with single `{`/`}`), similar to e.g. the xyl source url in OL
- is it really necessary to duplicate query, key and limit (endpoint plus additional attributes) ? the limit and the key could/should be part of the endpoint url (as these could have different names for other services), just the `query` parameter is set dynamically . It should probably be the other way round, that an attribute defines the `query` param name, meaning which url param to add if a search is made. Then it would be also compatible with other search endpoints
- then also handlebars is not needed any longer, in total it makes the whole element simpler

silvester-pari (approved):

Issue #570, opened 2024-01-11
Add view animation convenience functionality


silvester-pari:
E.g.
- zoom to coordinates/zoom
- zoom to geoJSON feature (passed as feature)
- zoom to rendered feature via vector/vector tile layer

santilland:
Please also consider zoom to extent

silvester-pari:
This could be an `animateView` property / `animate-view` attribute (I'd say default to `true`). If this is set, setting the `center` and/or setting the `zoom` would animate instead of "just" setting it.
It would need to check in the `_changedProperties' (`updated` hook) if both of them changed at the same time and do the animation only once.

Regarding features, extent etc. I don't have a clear idea yet...

silvester-pari:
Following scenarios are thouthg of:
- setting zoom and/or center: animate view to given zoom/center
- setting extent: animate view to given extent (overrules any zoom or center settings)
- setting layer property `fitView: true`: animate view to layer extent (overrules anything set before)
- `selectInteraction.selectById([id], true/{fitOptions})`: animate the view to fit progammatically selected feature(s)

lubojr:
Just a minor consideration of setting layer property `fitView: true` - how would it behave if more layers have this property? First found or the one which was freshly updated?

santilland:
The last option is once a feature is already loaded in a vector layer, it is possible to zoom to the available features?
I guess at least for stac we usually will have a bbox for the collection or item, so the extent should be ok for those cases.
I can't think of any more options right now, not sure if being able to pass a feature would make sense.

RobertOrthofer:
Two things are not fully implemented yet:

1.) selectById
we do have the method to `highlightById`, which basically is `selectById`, and we also have the `panIntoFeature` option on the select interaction. However, we don't have both together yet.

2.) `fitView` layer property
I'm not sure if we really want this option on the layer, as multiple layers will interfere with this and having an extent on the eoxMap and an overruling extent on the layer seems counter-intuitive. I think it would be best if the application could just control the `zoomExtent` property

silvester-pari:
Closed in favor of the two newly created issues.

PR #922, opened 2024-05-07
feat: allow map animations via properties

RobertOrthofer:
#570 

- [ ] All checks have passed

### DIFF
--- a/elements/map/helpers.ts
+++ b/elements/map/helpers.ts
@@ -239,3 +239,11 @@ export function registerProjection(
   proj4.defs(name, projection);
   register(proj4);
 }
+
+/**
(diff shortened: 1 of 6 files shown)


RobertOrthofer:
@silvester-pari this PR breaks some tests of the other elements, since there are animations now by default simple synchronous tests for zoom/center might not give the desired value.
We can either remove the default behaviour of "animate when zoom/center changes", or just set the property `animateOptions` in the tests to an empty object or `false`

silvester-pari:
@RobertOrthofer this would mean a breaking change, meaning we would need to bump the element to v2. Can we turn off the behaviour by default? And only enable it with a property/flag.

PR #956, opened 2024-05-15
chore: revert non-working parts of config extraction

silvester-pari:
## Implemented changes

The previously implemented functionality relied on a JS feature which turns out not to be reliable. Since the previous version (1.8.0) can't be unpublished any longer, it will be marked as "deprecated", and this change will be released as 1.8.1 instead.

- [x] All checks have passed


santilland (approved):
Thank you for the revert, this is the best way forward now, lets merge

PR #961, opened 2024-05-16
fix: prevent reaction to undefined zoom

silvester-pari:
## Implemented changes

This PR introduces a small check to the `zoom` setter if the value is actually defined, as setting `undefined` resulted in a blank map.

- [x] All checks have passed


RobertOrthofer (approved):
looks good to me, can't think of any real side effect right now...

PR #1006, opened 2024-06-05
feat: format options

### LINKED ISSUE #999: Extend source format to allow providing projection information for Vector data
Currently we can only pass the format type as string, eg.g. "format": "GeoJSON".
Some GeoJSON data soruces do not use expected 4326 projection, we need a way to pass a projection.
After some discussion with @silvester-pari we think allowing to pass also an object might be the best approach, e.g.:

```
"format": {"type" "GeoJSON", "options": {...GeoJSON options...}}
```
It would be good to understand if we can do this for other potential formats (GML, GPX, IGC, KML, MVT, WFS, WKB, WKT, XML, GPX, IGC, KML, WFS, WKT, GML2, GML3, GML32, OSMXML)?

 VectorTiles "layer" have their own projection attribute, so there we can already pass a projection?

The example we would have is:

```
"format": {
    "type" "GeoJSON", 
    "options": {
        "dataProjection": "'EPSG:3035'"
    }
}
```

RobertOrthofer:
This PR implements the possibility to use format options, additionally to the simple syntax that was implemented before. If the format is supposed to use a projection, that projection has to be registered via `eoxMap.registerProjection(...)` beforehand, like this has already to be done for view projections.

Because ol formats are very diverse in their appearance, i have left typing to a minimum, only requiring the "type" (like `GeoJSON`), and optional properties `dataProjection` and `featureProjection`, as i suppose that these will be the most used ones. Format options vary widely, `@ts-ignore`s will be needed for more exotic formats.

example simple format (unchanged):

```JSON
"source": {
      "type": "Vector",
      "url": "https://openlayers.org/data/vector/ecoregions.json",
      "format": "GeoJSON",
    },
```

example format with options:

```JSON
"source": {
      "type": "Vector",
      "url": "https://openlayers.org/data/vector/ecoregions.json",
      "format": {
            "type": "GeoJSON",
            "dataProjection": "EPSG:53009"
      }
    },
```

- [x] All checks have passed

closes #999


santilland:
I can use the `eoxMap.setAttribute("projection", "...");` to set correctly the projection for the entire map i have registered.
But when i use the format dataprojection parameter for a vector source it does not seem to be applied at all, not sure what the issue is, any ideas @RobertOrthofer ?

santilland (approved):
After integration testing and changing how the parameters are passed all seems to work as expected, so let's merge, thank you!

PR #1056, opened 2024-06-25
fix: order of format, layer and source spread

silvester-pari:
## Implemented changes

This PR changes the order of spreading the formats, layers and sources from the "advanced" plugin to happen first, and then the basic ones afterwards, as there were some issues that the internal ones were modified but then overwritten by the plugin ones.

- [x] All checks have passed


santilland (approved):
It seems to work also in eodash! Thank you! Let's merge!

Issue #587, opened 2024-01-16
Unmanaged style layer disappears when sorting


StefanBrand:
1. Feature is selected
2. Layer is dragged below other layer
3. Feature does not appear selected anymore

In fact, also the hover style is gone.

[Peek 2024-01-16 17-09.webm](https://github.com/EOX-A/EOxElements/assets/8948865/180175b2-0dbe-45be-95c1-cc76ed49fdcf)

silvester-pari:
I feel like this is a general issue with reordering if we have temporary layers. It would be nice to see if it happens with "vanilla" OL manipulation in order to make sure it's not a bug in layercontrol.

silvester-pari:
I managed to reproduce this issue in a codesandbox: https://codesandbox.io/p/sandbox/temporary-layers-2ccxlx
The issue is not connected to the layercontrol, so I'll go ahead and remove the layercontrol label.

silvester-pari:
@RobertOrthofer I don't get why the [`removeAt`](https://openlayers.org/en/latest/apidoc/module-ol_Collection-Collection.html#removeAt) and [`insertAt`](https://openlayers.org/en/latest/apidoc/module-ol_Collection-Collection.html#insertAt) functions should be a non-API-way of removing/adding layers from the viewpoint of ol?  :thinking: How else could we do the layer sorting in vanilla ol? I am aware of `map.removeLayer`, but it doesn't work in groups, and how to re-add the layer after sorting is done?

Having said that, I can't reproduce the sandbox code issue either in `1.2.0`, it seems that the interaction is not removed together with the layer any more (even with `map.removeLayer`, and even when using the eox-map `layers` property), so hover etc. still there even if the layer is gone. In `1.1.0` it removes the layer + interaction, but it does not come back after re-inserting the layer.

silvester-pari:
Fixed in #1078

PR #1078, opened 2024-07-03
Fix: unmanaged Styles

RobertOrthofer:
## Implemented changes

This PR handles the issue described in  #587. The idea is that sorting the ol layers by removing and re-adding them will not actually remove the connected interactions.

@silvester-pari is this how the desired behaviour? 

- [x] All checks have passed

### DIFF
--- a/elements/map/main.ts
+++ b/elements/map/main.ts
@@ -213,6 +213,16 @@ export class EOxMap extends LitElement {
           )
         ) {
           const layerToBeRemoved = getLayerById(this, l.properties?.id);
+          const jsonDefinition = layerToBeRemoved.get("_jsonDefinition");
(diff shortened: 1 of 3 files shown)


silvester-pari (approved):
Thank you, looks good to me, note that when selecting (& styling) a feature and the layer order is changed and the vector layer is underneath another layer, the styling for the feature is still applied. But this has always been the case and is expected I guess. From my side, good to merge!

RobertOrthofer:
That's true, I think this might also be the most practical solution from a cartographic standpoint for most real-world problems. I'm thinking that when changing the layer order, the selection layer most often is not dragged below a full opaque image-baselayer, but most likely underneath overlay layers or border layers, where the geometry of the selected feature "is expected" to stay on top

PR #1100, opened 2024-07-17
feat: enable/disable compare

silvester-pari:
## Implemented changes

This PR adds a new functionality to the `eox-map-compare` and allows to toggle the comparison mode via the new `enabled` attribute:
- `enabled="first"` only shows the first map
- `enabled="second"` only shows the second map
- `enabled` (`undefined` or any other truthy value; default) shows both maps

This effectively allows enabling/disabling the comparison in a flexible way.

- [x] All checks have passed

### DIFF
--- a/elements/map/map.stories.js
+++ b/elements/map/map.stories.js
@@ -512,6 +512,9 @@ export const TooltipWithPropertyTransform = {
   `,
 };
 
+/**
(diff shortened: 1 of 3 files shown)


santilland (approved):
I really like this approach, it allows us to enable/disable comparison, but also easy implementation of a compare toggle button instead of a slider, from my side happy if this is merged!

PR #1106, opened 2024-07-18
feat: allow passing an eox-map DOM element to `sync` property

silvester-pari:
## Implemented changes

This PR adds the functionality to pass an `eox-map` directly to the `sync` property. Previously it only took a `string` and internally did a `document.querySelector`, but if the `eox-map` is used within another shadow root, then this query selector didn't work.

Now there are two options:
- pass a `string` as attribute or property, e.g. `<eox-map sync="#other-map">`
- pass a DOM element as property, e.g. `eoxMapB.sync = document.querySelector("#other-map")` 

- [x] All checks have passed

### DIFF
--- a/elements/map/main.ts
+++ b/elements/map/main.ts
@@ -399,11 +399,32 @@ export class EOxMap extends LitElement {
     }
   }
 
+  private _sync: string | EOxMap | undefined;
   /**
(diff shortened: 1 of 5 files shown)


A-Behairi (approved):
LGTM. Thanks for adding the `getElement` utility function!

Issue #713, opened 2024-03-05
Item Filter Items should be deselectable


lubojr:
As a configurable feature I would like the ItemFilter items to be deselectable as well.
As a default current behavior (single select without option to deselect) is fine.

silvester-pari:
We currently use radio inputs for the result list, and I believe there is no way to deselect a radio input. We could consider using something else instead of radio buttons.

santilland:
@silvester-pari is it possible to deselect items now that radio buttons are no longer used?

silvester-pari:
Yes, now results can be selected and deselected

PR #1109, opened 2024-07-18
feat: improved list style

silvester-pari:
## Implemented changes

This PR introduces a new style for result lists:

- [x] removed radio input
- [x] deselectable results by default (https://github.com/EOX-A/EOxElements/issues/713)
- [x] `idProperty` analogous to `titleProperty` and highlighting checks based on that (see https://github.com/EOX-A/EOxElements/issues/835 - if an item is really rendered two times because it is aggregated in two lists, then both instances are highlighted)
- [x] optional `subTitle` property

## Screenshots/Videos

image
image
image

- [x] All checks have passed

### DIFF
--- a/elements/itemfilter/src/components/filters/selector.js
+++ b/elements/itemfilter/src/components/filters/selector.js
@@ -217,9 +217,6 @@ export class EOxSelector extends LitElement {
               <li
                 data-identifier="${suggestion.toString().toLowerCase()}"
                 data-title="${suggestion}"
-                class="${this.selectedItems.includes(suggestion)
-                  ? "highlight-item"
-                  : ""}"
               >
                 <label>
                   <input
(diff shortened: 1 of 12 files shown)


santilland:
I think removing the radio button from the result list is a good idea!
What for me feels a little "conflicting" is that "selecting" a result looks like activating one of the filter entries. I think i would remove the background highlight for activated filter entries. For me these are two different "actions" which should not look equal.
Another option might be two different highlight color? Not sure i like that option.
Looking at the first screenshot for me the selected result which in a sense is the "final action" intended for the itemfilter is not distinguished enough.

silvester-pari:
> I think removing the radio button from the result list is a good idea! What for me feels a little "conflicting" is that "selecting" a result looks like activating one of the filter entries. I think i would remove the background highlight for activated filter entries. For me these are two different "actions" which should not look equal. Another option might be two different highlight color? Not sure i like that option. Looking at the first screenshot for me the selected result which in a sense is the "final action" intended for the itemfilter is not distinguished enough.

I now tried removing the highlighting of the selection filters (multiple and single select), I think it is now even cleaner:

image

PR #1132, opened 2024-07-25
chore: allow node versions 20.x

silvester-pari:
## Implemented changes

As it turns out, the changes made in https://github.com/EOX-A/EOxElements/pull/1108 where not a smart idea. Pinning the exact version caused some troubles, e.g. in the Netlify pipeline (which only allows to specify `20.x` and now started using the new LTS `20.16.0`. In order to not have to adjust each time a new LTS is specified, `20.x` should be more robust.

- [ ] All checks have passed

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -35,7 +35,7 @@
         "storybook": "^8.0.0"
       },
       "engines": {
-        "node": "20.15.1"
+        "node": ">=20.0.0 <21.0.0"
       }
     },
     "elements/chart": {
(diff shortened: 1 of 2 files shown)


silvester-pari:
As just discussed, this might also not be the best approach as this would possibly restrict the usage of our elements

silvester-pari:
Merging even though the tests fail (see other issue with ol-dev), but at least the other pipelines should start working again now.

Issue #975, opened 2024-05-21
smart layer update for changes not working for grouped layers


santilland:
The smart update functionality for layers assumes a flat structure, but it is possible to work with groups.
We either need to rethink how we do the smart layer update approach, or implement recursive checks for potentially any depth level that is possible with groups

RobertOrthofer:
What exactly is not working? I tried changing the opacity and style 2 levels deep, that seems to work just fine

santilland:
Hello @RobertOrthofer , as far as i can see, if the group identifier stays the same, but the nested group layers change, with new ids and everything the nested layers will not update.
So for example i created 3 groups, overlays, baselayers and data layers, with "constant" ids.
If i load a new layer config, with these groups, but different nested content, the content of the group will not be updated correctly.
I solved it for now by generating a pseudo uid for the layer groups each time i make a change in the nested layer content.

RobertOrthofer:
Thanks for the detailed test case, i will look into it

santilland:
I have continued testing and as you said, it seems to work as expected, i can't reproduce the issue again right now.
I have found another issue testing where changing layer config after a webGL layer or vectortile layers was added produces an error trying.
Somehow there is an undefined when doing the forEach here:
https://github.com/EOX-A/EOxElements/blob/974aa84ee81439b5e5fdf8dca9caf03742bf88ec/elements/map/src/generate.ts#L400

Not sure if i should open a new issue, or we can leave it as part of this one.

Here are the configs:
```
[...]

PR #1147, opened 2024-07-29
fix: layer update of group layers

RobertOrthofer:
This PR fixes a bug when updating layers inside layer groups.

Additionally, this PR adds another test to more realistically simulate real-world situations when working with groups. This test also detects failing `instanceof`-checks that occured in the past when multiple instances of `ol` are installed in the project.

connected to https://github.com/EOX-A/EOxElements/issues/975 

- [x] All checks have passed

### DIFF
--- a/elements/map/src/generate.ts
+++ b/elements/map/src/generate.ts
@@ -396,7 +396,7 @@ export function updateLayer(
     const layerCollection = (
       existingLayer as unknown as import("ol/layer/Group").default
     ).getLayers();
-    layerCollection.forEach((l: Layer) => {
+    layerCollection.getArray().forEach((l: Layer) => {
       if (!newLayerIds.includes(l.get("id"))) {
         layerCollection.remove(l);
       }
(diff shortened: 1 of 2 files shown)


RobertOrthofer:
If this is ok, i think we can close #975

silvester-pari (commented):
Looks good from my side, but I'm not very advanced in this topic, so I'd like to get a review from @santilland on this.

santilland:
I updated i think all packages, but i am still getting the error in https://github.com/EOX-A/EOxElements/issues/975, maybe i oversaw something? I still get the same error in the current story deployment here:
https://eox-a.github.io/EOxElements/?path=/story/elements-eox-map--geo-tiff-layer
Any pointers? @RobertOrthofer

silvester-pari:
You're right, I tried on https://deploy-preview-1147--eoxelements.netlify.app/?path=/story/elements-eox-map--geo-tiff-layer and still get the error as described in #975 , with the error pointing to https://github.com/EOX-A/EOxElements/blob/map/test/grouplayer-instanceof/elements/map/src/generate.ts#L400
image

santilland (approved):
That was a combination of things that made finding this quite complicated, dev deployment works with the example in the issue, very much in favor of merging the fix, thanks for the investigation to all :)

PR #1152, opened 2024-07-29
feat: allow passing `eox-map` DOM element to `for` property reactively

spectrachrome:
## Implemented changes

Enables reactive updates to `for` properties for eligible elements with the help of the `update` Lit lifecycle hook and getter/setter reactivity wrapping `this.#eoxMap` in `this.eoxMap`.

| Element  | Eligibility | Reactive?
| ------------- | ------------- | ------------- |
| `eox-drawtools`  | ✅ Yes  | ✅ Implemented
| `eox-geosearch`  | ✅ Yes   | ✅ Implemented
| `eox-layercontrol` | ✅ Yes  | ✅ Implemented
| `eox-timecontrol` | ✅ Yes  | ✅ Implemented
| `eox-stacinfo`  | ⛔️ `for` attribute only takes URL strings | 
| `eox-itemfilter` | ⛔️ No `for` attribute |
| `eox-jsonform`  | ⛔️ No `for` attribute  |
| `eox-chart`  | ⛔️ No `for` attribute  |
| `eox-layout`  | ⛔️ No `for` attribute  |

Because of these changes, the workaround used in eodash_v5 is no longer necessary, as the `for` attribute can be used directly like this:

```vue
<template>
  <span class="d-flex flex-column fill-height overflow-auto pa-4">
    <eox-drawtools :for="eoxMap.value" />
  </span>
</template>

<script setup>
import "@eox/drawtools";
import { onMounted, ref, computed } from "vue";
import { store } from "@eodash/eodash";

const eoxMap = computed(() => store.states.mapEl);
</script>
```

_The name of this branch has been changed from `main/feature/element-for` to `all/feature/element-for` as Git won't allow a sub-branch starting with `main`._

- [ ] <s>For new features: I have created a story showcasing this new feature</s>[^1]
- [x] All checks have passed

[^1]: At the moment, it is not possible to test this functionality within Storybook, as the use of a `document.querySelector('eox-map')` in the `for` attribute will trigger a cyclic reference error. I was not able to determine exactly where this cyclic value occurs. I used my local test instance of `eodashv5` and compiled bundles to verify that the implementation is working.


silvester-pari (changes requested):
Thank you for this PR and especially for the detailed description! From a technical point of view it looks good, however it could be a good moment in time to also harmonize the docs/types accordingly. In Some elements we define the type as only `String`, whereas in others as `String|HTMLElement` - see e.g. https://github.com/EOX-A/EOxElements/blob/a053bbb54b31170fff8604f173f6523410c797d6/elements/layercontrol/src/main.js#L72

This affects the automatically generated Docs, e.g.: 
image

I propose to harmonize this in all elements and write something like:

```js
    /**
     * Query selector of an `eox-map` (`String`, passed as an attribute or property)
     * or an `eox-map` DOM element (`HTMLElement`, passed as property)
     *
     * @type {String|HTMLElement}
     */
    this.for = "eox-map";
```
(plus, making sure in all elementss "eox-map" is the default value).

Result: 
image

silvester-pari (approved):

PR #1161, opened 2024-07-30
fix: issue of layer groups not removing all previous layers

santilland:
## Implemented changes

This PR attempts once more to fix the layer group update by creating a shallow copy before iteration, so that removing an item from the original array doesn't affect the logic negatively.

- [x] All checks have passed

### DIFF
--- a/elements/map/src/generate.ts
+++ b/elements/map/src/generate.ts
@@ -396,7 +396,12 @@ export function updateLayer(
     const layerCollection = (
       existingLayer as unknown as import("ol/layer/Group").default
     ).getLayers();
-    layerCollection.getArray().forEach((l: Layer) => {
+    // We can't remove the layers of a collection while iterating on it,
+    // this fails in removing all layers as expected.
+    // One solution is to create a shallow copy which we iterate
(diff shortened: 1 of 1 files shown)


RobertOrthofer:
This single line of code really keeps on giving. IMO, the cleanest way would be a reverse for-loop, like

```JavaScript
for (let i = arr.length - 1; i >= 0; i--) {
  // ...
}
```

Or we don't manipulate the array in the loop at all, only store the references of of items to be removed, and call `collection.remove(item)` for each of those items.

santilland:
I was not sure what to consider when manipulating the underlying collection, so i kept it as "close to the original" as possible. Also fine to populate a "to delete list".
I was also thinking of the reverse iteration, but as we do not work with index, i could foresee potential re-ordering issues while iterating and deleting. If we reverse the iteration then we should work with removing the layers based on the index i would propose?
A shallow copy seemed like the safest bet, but feel free to modify as you think fit.

silvester-pari (approved):

RobertOrthofer (approved):
You are right, I think it is fine as is.

Issue #1097, opened 2024-07-16
Select By ID


RobertOrthofer:
we do have the method to `highlightById`, which basically is `selectById`, and we also have the `panIntoFeature` option on the select interaction. However, we don't have both together yet.

_Originally posted by @RobertOrthofer in https://github.com/EOX-A/EOxElements/issues/570#issuecomment-2100797221_

PR #1176, opened 2024-08-02
feat: highlight and animate feature(s) by ID

RobertOrthofer:
This PR implements animations for programmatic feature selection for both VectorLayer and VectorTileLayer without causing a version jump of the map element.

The internally used `panIntoFeature` function has been lifted to the level of the interaction, exposing it the same way as the `highlightById` function as a method of the interaction itself.

The existing `highlightById`-method now has an optional second parameter, `fitOptions`, allowing animations to highlighted Features of `VectorLayer` and RenderFeatures of a `VectorTileLayer`. Be aware that when programmatically selecting a feature, the source has to be loaded, and, in the case of the `VectorTileLayer`, the feature has to be available in currently rendered tiles. The [`fitOptions`](https://openlayers.org/en/latest/apidoc/module-ol_View.html#~FitOptions) also include convenient properties like `padding` or `maxZoom`.

This is designed as convenience-functionality for the `select`-Interaction, as an improvement of the existing `highlightById` method. Arguments could be made for the `panIn`-flag to be replaced completely by `fitOptions`, in order to have more control over animations (this is currently static with `duration: 750` and no padding), but this would mean a version jump.

#570
#1097 

- [x] All checks have passed

### DIFF
--- a/elements/map/map.stories.js
+++ b/elements/map/map.stories.js
@@ -513,6 +513,67 @@ export const TooltipWithPropertyTransform = {
 };
 
 /**
+ * Select interactions offer a `highlightById` method, with which vector features can be programmatically selected via their id property.
+ * It expects an array with a list of ids to be selected.
+ * Optionally, passing a second argument allows to set the [`fitOptions`](https://openlayers.org/en/latest/apidoc/module-ol_View.html#~FitOptions),
+ * adding view animation to the selection.
+ */
+export const ZoomToFeatures = {
+  args: {
+    config: {
+      layers: [
+        {
+          type: "Vector",
+          background: "lightgrey",
+          properties: {
+            id: "regions",
+          },
+          source: {
+            type: "Vector",
+            url: "https://openlayers.org/data/vector/ecoregions.json",
+            format: "GeoJSON",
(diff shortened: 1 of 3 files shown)


silvester-pari (changes requested):
Great addition, thanks! Fine for me to add this now, and add an issue that deprecates the `panIn` property for v2.x

Could we just add a story for this functionality? I'll see if I find time for this today, otherwise any other volunteers welcome :smiley:

silvester-pari:
I added a story showcasing this:

[Screencast from 2024-08-05 15:01:52.webm](https://github.com/user-attachments/assets/65e2cd04-d6cb-48f2-a642-c6ed98be626f)

santilland (approved):
For me also looks good, not using it right now in any application, so probably good to see if @StefanBrand has feedback?

RobertOrthofer:
thanks, this story is indeed a very good showcase of this feature

RobertOrthofer:
About the `panIn`-option, maybe we don't need to get rid of it entirely, as it is a good way to pan in to something without using JS (only via the config), but we could use `fitOptions` instead of a `boolean` flag

silvester-pari (approved):

PR #1181, opened 2024-08-05
feat: lonLatCenter & lonLatExtent

RobertOrthofer:
## Implemented changes

This PR adds the read-only-properties `EOxMap.lonLatCenter` and `EOxMap.lonLatExtent`, which return the current center and extent in EPSG:4326, no matter the current projection of the view. 
Additionally, the new methods `transform` and `transformExtent` are now exposed, allowing more flexible transformations between all registered coordinate systems. Basically, these methods are the `ol/proj` [transform](https://openlayers.org/en/latest/apidoc/module-ol_proj.html#.transform) and [transformExtent](https://openlayers.org/en/latest/apidoc/module-ol_proj.html#.transformExtent) functions, with the small adaptation that the `destination` defaults to `EPSG:4326` if not defined.

- [x] All checks have passed

### DIFF
--- a/elements/map/main.ts
+++ b/elements/map/main.ts
@@ -125,6 +125,29 @@ export class EOxMap extends LitElement {
     return this._center;
   }
 
+  /**
+   * current center of the view in EPSG:4326
+   */
+  get lonLatCenter() {
+    if (this.projection === "EPSG:4326") {
+      return this.map.getView().getCenter();
+    }
+    return transform(this.map.getView().getCenter(), this.projection);
+  }
+
+  /**
+   * current extent
+   */
+  get lonLatExtent() {
+    const currentExtent = this.map
+      .getView()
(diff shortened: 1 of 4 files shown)


silvester-pari (changes requested):
Again, here it would be nice to have a story showcasing this.

Plus, some description/documentation for the API docs: 
image

silvester-pari:
I added a **very** basic story showcasing this, but probably better than nothing.

silvester-pari (approved):

RobertOrthofer:
Thanks for the story. We should do documentation of all utility functions in the same place, once they are all exported. They probably don't need a map in their story

santilland (approved):
Initial integration tests showed a small issue that was corrected, further tests seem to work as expected, so i think this is good to go!
Thank you @RobertOrthofer for the implementation!

PR #1173, opened 2024-08-02
feat: loading indicator map control

RobertOrthofer:
## Implemented changes

This PR implements a Loading Indicator Control for the map element as described in #911, providing a default SVG graphic.

possible options for this control:

 -  `type`: `"small"` or `"fullscreen"` (defaults to `"small"`)
 - `opacity`: opacity of the spinner from 0 to 1
 - `spinnerSvg`: optional SVG as markdown as string, should already provide the desired animation (`"<svg xmls...  </svg>"`)

Two stories have been added, because the different types look quite different. Most of the time the "small" type should be sufficient.

## Screenshots/Videos

https://github.com/user-attachments/assets/0e7c86dd-12f0-4867-a7da-95cf694dc394

https://github.com/user-attachments/assets/22e7fc55-ec0a-41b1-8010-f931b72cbc42

## Checklist before requesting a review

- [x] I have performed a self-review of my code
- [x] I have added a test related to this feature/fix
- [x] For new features: I have created a story showcasing this new feature
- [x] All checks have passed
- [x] The PR title [follows conventional commit style and reflects the type of change (major/minor/patch, breaking)](https://github.com/googleapis/release-please?tab=readme-ov-file#how-should-i-write-my-commits)
- [x] The PR title describes the change within this element (each merged PR equals one entry in the element's changelog!)

PR #1200, opened 2024-08-07
feat: implement loading indicator

spectrachrome:
## Implemented changes

Integrates the recently implemented standard map loading indicator into the geosearch element, with an option to customize the SVG loader using the `loaderSvg` attribute, as demonstrated in the new Custom Loader story.

Do you, @silvester-pari think maybe we should centralize the loader/spinner SVG, similar to how we handled the styles for buttons, radio buttons and checkboxes? Also I think maybe a blue would fit better with the rest, instead of orange.

### Default

Screen Recording 2024-08-07 at 14 03 01

### Custom Loader

Screen Recording 2024-08-07 at 14 21 43

- [x] All checks have passed

### DIFF
--- a/elements/geosearch/src/main.js
+++ b/elements/geosearch/src/main.js
@@ -32,6 +32,11 @@ class EOxGeoSearch extends LitElement {
        */
       _query: { attribute: false },
       /**
+       * Returns true if the element is currently loading data from the API.
+       * @private
+       */
+      _isLoading: { attribute: false },
+      /**
        * The OpenCage API endpoint to use for the search, including the key but without the query parameter.
        *
        */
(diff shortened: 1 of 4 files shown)


silvester-pari (changes requested):
Thank you, yes we could think of centralizing the default spinner to a shareable package.

But the bigger concern currently is that the spinner is hardcoded in your proposed change, so there is no way to replace/style it. Usually, we either provide CSS variables or dedicated properties for customizing such things.

Also, it seems you didn't use the default spinner from eox-map, maybe you used the one from an example story? I think the default one is much thinner and not colored red.

spectrachrome:
Yes, I used the full screen variant of Robert's loading indicator:
https://github.com/EOX-A/EOxElements/pull/1173/files#diff-d6260890d082a80ce9b6cae725a4300e6e6f74e3aa3b27090b11aa189359e83cR360-R362

Thank you for the review, I will try to use the normal map loader and use either CSS variables or something like a custom `loaderSvg` attribute to make it customizable.

spectrachrome:
Default spinner is now used, an alternative SVG spinner can be provided in the `loaderSvg` attribute and a new Custom Loader story has been added.

silvester-pari (approved):

PR #1205, opened 2024-08-08
feat: added optional extent for the registerProjection function

santilland:
## Implemented changes

This PR adds `extent` as an optional third parameter to the `registerProjection` function. This allows setting the validity extent for the set projection. Details see [the corresponding OL docs](https://openlayers.org/en/latest/apidoc/module-ol_proj_Projection-Projection.html#setExtent).

- [x] All checks have passed

### DIFF
--- a/elements/map/helpers.ts
+++ b/elements/map/helpers.ts
@@ -239,13 +239,12 @@ export async function registerProjectionFromCode(code: string | number) {
 export function registerProjection(
   name: string,
   projection: string | proj4.ProjectionDefinition,
-  extent?: number[]
+  extent?: import("ol/extent").Extent
 ) {
   proj4.defs(name, projection);
   register(proj4);
   if (typeof extent !== "undefined") {
-    const projection = getProj(name);
-    projection.setExtent(extent);
+    getProj(name).setExtent(extent);
   }
 }
 
--- a/elements/map/helpers.ts
+++ b/elements/map/helpers.ts
@@ -9,6 +9,7 @@ import { getArea, getLength } from "ol/sphere";
 import { getUid } from "ol";
 import { DrawEvent } from "ol/interaction/Draw";
 import { DragAndDropEvent } from "ol/interaction/DragAndDrop";
+import { get as getProj } from "ol/proj";
 import Feature from "ol/Feature";
 import proj4 from "proj4";
 import { fromEPSGCode, register } from "ol/proj/proj4";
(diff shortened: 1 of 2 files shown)


lubojr (approved):
as per eodash v4 implementation, approve

elements/map/helpers.ts
```diff
@@ -237,10 +238,15 @@ export async function registerProjectionFromCode(code: string | number) {
  */
```
silvester-pari, line 238: please add some JSDoc description, e.g. `@param extent the validity extent for the projection`

elements/map/helpers.ts
```diff
@@ -237,10 +238,15 @@ export async function registerProjectionFromCode(code: string | number) {
...
 export function registerProjection(
   name: string,
-  projection: string | proj4.ProjectionDefinition
+  projection: string | proj4.ProjectionDefinition,
+  extent?: number[]
 ) {
   proj4.defs(name, projection);
   register(proj4);
+  if (typeof extent !== "undefined") {
+    const projection = getProj(name);
```
silvester-pari, line 247: `projection` is already defined as function param, could we use another variable name here?

elements/map/helpers.ts
```diff
@@ -237,10 +238,15 @@ export async function registerProjectionFromCode(code: string | number) {
  */
 export function registerProjection(
   name: string,
-  projection: string | proj4.ProjectionDefinition
+  projection: string | proj4.ProjectionDefinition,
+  extent?: number[]
```
silvester-pari, line 242: In other places of `eox-map`, we have defined the type for extent as follows:
```js
  extent?: import("ol/extent").Extent
```
This also fits better with the `setExtent` function which expects an `Extent`.

silvester-pari (changes requested):
Thanks for this addition, please add some description to the PR, and ideally we would also include the extent in a test. I'll see if I can prepare something in this regard!

elements/map/helpers.ts
silvester-pari: Or maybe it is enough to just do `getProj(name).setExtent(extent)`?

santilland: good points, i updated it accordingly

elements/map/helpers.ts
santilland: Also updated extent type definition, thanks!

PR #1242, opened 2024-09-07
chore: fix flaky tests

radupasparuga:
## Implemented changes
Some tests (e.g. `map/viewProjection`) have assertions that fail after the test has ended.
Screenshot 2024-09-09 at 14 16 58
In the above screenshot, the `lonLatExtent` assertion is actually part of "use special projection" test but because of a `setTimeout` it gets pushed down to the next test.
In order to fix this we wrap the assertion in a `Cypress.Promise` (https://docs.cypress.io/api/utilities/promise) as we do in other tests that require a `setTimeout` and as a result the test now fails which needs fixing on the map side.
Screenshot 2024-09-09 at 14 24 31

- [ ] All checks have passed

### DIFF
--- a/elements/map/test/viewProjection.cy.ts
+++ b/elements/map/test/viewProjection.cy.ts
@@ -97,98 +97,101 @@ describe("view projections", () => {
   });
 
   it("use special projection", () => {
-    cy.intercept(
-      "https://openlayers.org/data/vector/ecoregions.json",
-      (req) => {
-        req.reply(ecoRegionsFixture);
-      }
-    );
-    // not using osm because of performance issues while testing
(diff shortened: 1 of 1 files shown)


RobertOrthofer:
I am not yet sure why this test is failing, but the getter from `lonLatExtent` is not asynchronous (anymore?), so there should be no need for a `setTimeOut` to begin with. I will look into this.

RobertOrthofer:
@radupasparuga I've extracted the failing test for clarity. It now uses the simpler `EPSG:32633` with a more reasonable testing extent, my best bet is that the `extent` of the definition of the polar projection was not correct, but the failing test wasn't awaited correctly.

RobertOrthofer:
thanks for the cleanup, i missed that.

silvester-pari (approved):

PR #1276, opened 2024-09-26
feat: format handler for multiple geometry types and formats

santilland:

- [x] All checks have passed


santilland:
@silvester-pari i think this brings great first implementation to support different types of geometries into the jsonform using the draw tools. There will be nuances that need to be solve, e.g. also to better support specialized STAC schemas, but i think as is it already provides great value and would propose to merge and continue improving in smaller increments afterwards.
Thank you @A-Behairi for the implementation!

silvester-pari (changes requested):
Thank you, looks like a big addition! I am unsure about a few points however:
- the `*-editor` formats: I believe it doesn't make sense to offer this distinction, as the drawing of a bbox should always allow to edit the coordinates manually (via a number input) in order to fine-tune the drawing. I get that the current editor that comes with drawtools is not ideal for that (layout, and showing the entire feature content instead of just the coordinates), but here we maybe can think of a different solution (e.g. keep the coordinate input fields as per default json-editor, and just add the map additionally)
- having to click the "draw" button in order to be able to start drawing: I feel like from a user perspective this is an unnecessary step and could cause quite some friction. I understand that this is the technical solution we have, but it's less than ideal from the user experience. Ideally the user comes to the section where the map appears (by default it should not be scroll-zoomable by the way) and can simply draw the corresponding shape
- changing of the current example schemas: it's probably not wise to change the current story schemas (as in the [collection schema](https://github.com/EOX-A/EOxElements/pull/1276/files#diff-9543913249cb73b04190d18de0a130a2b2cb72b357688f87c651295025110424L634)) just to showcase something that works quite differently; the current example schema is  designed with STAC in mind and I wouldn't like to completely dismiss this (when we remove it now from the codebase, will it ever come back?). I would have hoped that the new functionality would be an addition that works with the existing schema, and not require a completely new one.

All in all, it feels a bit like a very much technology-driven solution, not necessarily a developer- and user-friendly one; maybe we can find compromises, offering less functionality but in a more user-friendly manner? :thinking: 
Maybe also the drawtools are not the ideal fit for the jsonform, and we maybe should just use the map and then some input fields generated by jsonform; but happy to do some more brainstorming and/or experimentation!

A-Behairi:
Thank you for your feedback! I would like to note that the implementation was intended with much developer experience and user experience in mind, I would like to clear and comment on some points:

*  the implementation doesn’t change or interrupt any of the current supported schemas/editors, it builds on top of it by introducing `type: “spatial”` with various supported formats, options, and a custom validator.

> (when we remove it now from the codebase, will it ever come back?)

That being said, we can simply revert the change in the collection and it would go back to how it was.

* I can see that there’s alot of enhancements made to the current collection schema vs the actual stac collection spec prior to this PR (for example the license property). The changes that this PR introduced doesn’t misregard the STAC collection spec at all, the returned value from `{ type:”object”, format:”bounding-boxes }”` is an array of 4 element number arrays, which is matched by the spec.  

PS. the changes has been made after @silvester-pari asked in the daily about how would this PR be integrated with the collection schema, and it was intended to show case that. 

* the spatial editor has a title property that can guide the user to select/draw an area/feature on the map, please checkout the the image below as an example :
Screenshot 2024-11-15 at 11 01 23

I think this fulfills the requirement of it being UX friendly, but If you think that it would be better to start the drawing without pressing on the button, I would be happy to implement that.

* regarding the `*-editor` formats, I think this is intended for advanced users, I agree it could look cleaner and I think we can either include it now and iterate on it in the future or remove it and implement it as a separate feature in a more UI/UX friendly way.

silvester-pari:
Thank you @A-Behairi for the detailed answer! Here are my thoughts:

- `*-editor` formats, I agree with your suggestion to remove them completely for now and work on this in a future PR
- drawing without pressing button: how about we do a compromise, and say if it uses an external map (`for` property) then keep as it is, if it does not use `for` then enable drawing automatically?
- type `spatial`: I will try to explain better what I mean below:

This was the previous schema:
```
              "properties": {
                "bbox": {
                  "type": "array",
                  "items": {
                    "type": "array",
                    "format": "bounding-box",
                    "items": { "type": "number" },
                    "minItems": 4,
                    "maxItems": 4
                  },
                  "minItems": 1
                }
              }
```
For bounding boxes, I would imagine something like this:
```
              "properties": {
                "bbox": {
                  "type": "array",
                  "format": "bounding-boxes",
                  "items": {
                    "type": "array",
                    "format": "bounding-box",
                    "items": { "type": "number" },
                    "minItems": 4,
                    "maxItems": 4
                  },
                  "minItems": 1
                }
              }
```
Note a few things:
- on the "parent" level, the type is `array`, since it is expected to output an array. Only the format is `bounding-boxes`, which renders a map with multi-bbox input instead of an array editor
- on the "child level", the type is still `array` as it expects an array of coordinates as output, and the format is `bounding-box`, since it allows drawing a single bbox on a map
- if for example the "parent" would have no format specified (only the children), it would render an array editor, and each array would have its own bounding-box map. By providing `bounding-boxes` format on the parent, it allows combining all these in one map instead of having four maps
- there could be usecases where only a single `bounding-box` is needed
- the type `spatial` is not really describing an output format, so I'm not sure we need to have a type `spatial`. I hope it works to just implement a custom rendered for the type `array`

santilland:
hi! it seems there is quite some definition work that was needed for this, as it is bringing in concepts needed in multiple places, i will try to recap things to see if we agree on it (i already had a quick catch up with Ahmed about these points).
1. we agree to remove the `*-editor` cases. As i see it in the future we will want to have a dedicated field for the values, allowing also to edit them, some thoughts on this later
2. as for automatic drawing activation i think changing the behavior depending on the "scenario" can be very confusing for someone using the eox-json-form. My proposal would be we set the automatic start of selection handler as default and provide an option to disable starting the selection handler  (e.g. `autoStartSelection: false`)
3. changing the type to array makes sense for bbox and polygon but for feature it gets more complicated. More on that below.

Topic:
* fields for editing values: the default way to handle arrays in json form is to have dedicated fields for each value, this should be fine for bbox as it is only 4 values, for a polygon though it would be a crazy number of fields.
* changing type of feature(s): the idea with feature(s) is allowing selecting a vector feature and extracting a property of it, this can be defined in the options. This property could be many things, e.g. it's extent, identifier, name, geometry, .... I think we need to explore if we can use a feature format for a generic type, that might be one possible approach

Additionally:
After talking use cases with Elias i realized we need also point selection. I saw draw tools supports this, so i would propose to call the format `point`? Would also be of type array, should be very aligned to bbox 2 values instead of 4.

santilland:
Trying to integrate a new statistical api experiment i realized that geometries can be expected in at least a couple of different ways.
In order to avoid needing strange conversion in the client i think we need to support these types, and thinking about how to solve this maybe we can nicely use the type to solve this.
Currently i see 3 different approaches:
* geojson: geojson of the geometry
* WKT: WKT representation of the geometry
* array: array of the geometry (if multiselection allowed array of arrays)
  - point: 2 values
  - bbox: 4 values 
  - polygon: N values

santilland:
Hello @A-Behairi this is looking great, the type options are really helpful to support different endpoint capabilities.
Trying to integrate one endpoint i realized that the geojson type is a little more complex as it allows many ways to structure it. I think we can in principle use the templating language to extract exactly what we need but it would be good to evaluate how we structure the geojson. 
I would propose instead of using featurecollections for all, to use simple features for the single selection cases, i will try to add some examples below. 
For the multiselect (points, bounding-boxes, polygons) it would make more sense to have feature collections, although again, reading about options it could also be possible to use [MultiPoint](https://geojson.org/geojson-spec.html#id5) [MultiLineString](https://geojson.org/geojson-spec.html#id6) and [MultiPolygon](https://geojson.org/geojson-spec.html#id7). I think i like this option more as it ensures all geometries are from the same type? In a feature collection there could potentially be any mix of things.

Any opinion on this would be very welcome, please let me know what you think.

While writing this i realized i know of use cases where we will need to define a line, so if it is not too much effort we could use the opportunity to add this format also here, we could call it `format: line` and `format:lines`? I added an example also at the end. 

Examples for single select:

{type: "geojson", format: "point"}

```
{
      "type": "Feature",
      "geometry": {
        "type": "Point",
        "coordinates": [
          11,
          43
        ]
      },
      "properties": {}
    }
```

{type: "geojson", format: "polygon"} and {type: "geojson", format: "bounding-box"}

```
{
      "type": "Feature",
      "properties": {},
      "geometry": {
        "coordinates": [
          [
            [
              11.186767387616754,
              44.019851231386326
            ],
            [
              11.230712935404142,
              42.17497898908604
            ],
            [
              12.428222447740097,
              41.9629345529863
            ],
            [
              13.988281347309453,
              42.734276566185855
            ],
            [
              13.39501999765912,
              44.2012799811516
            ],
            [
              11.186767387616754,
              44.019851231386326
            ]
          ]
        ],
        "type": "Polygon"
      }
    }
```

{type: "geojson", format: "line"}

```
{
      "type": "Feature",
      "properties": {},
      "geometry": {
        "coordinates": [
          [
            12.086527294554173,
            43.151923751754
          ],
          [
            13.318383515919493,
            43.141078166340606
          ],
          [
            13.790678332841026,
            43.17633857773657
          ]
        ],
        "type": "LineString"
      }
    }
```

elements/jsonform/stories/public/collectionSchema.json
```diff
@@ -632,14 +632,7 @@
...
                   "type": "array",
-                  "items": {
-                    "type": "array",
-                    "format": "bounding-box",
-                    "items": { "type": "number" },
-                    "minItems": 4,
-                    "maxItems": 4
-                  },
-                  "minItems": 1
+                  "format": "bounding-boxes"
```
silvester-pari, line 635: One small adjustment after some research: the `bounding-boxes` format would move to this "level", as you correctly intended, and is therefore not needed any longer in line 630 (one level "higher"). The rest of the schema should probably stay the same as before, following the (simpliefied) [STAC JSON schema](https://github.com/radiantearth/stac-spec/blob/ec002bb93dbfa47976822def8f11b2861775b662/collection-spec/json-schema/collection.json#L69). So, in total the `spatial` property would look like this:

```json
            "spatial": {
              "title": "Spatial Extents",
              "type": "object",
              "required": ["bbox"],
              "properties": {
                "bbox": {
                  "type": "array",
                  "format": "bounding-boxes",
                  "items": {
                    "type": "array",
                    "format": "bounding-box",
                    "items": { "type": "number" },
                    "minItems": 4,
                    "maxItems": 4
                  },
                  "minItems": 1
                }
              }
            },
```

This renders nicely the map with multi select and correctly sets the values!
image

silvester-pari (approved):
Thank you for this massive addition! :tada:

PR #1307, opened 2024-10-08
feat: charts using Vega-Lite

silvester-pari:
## Implemented changes

This introduces [Vega](https://vega.github.io/vega/) as a basis for `eox-chart`.
The basic chart functionality is based on [Vega-Lite](https://vega.github.io/vega-lite/)/[Vega-Embed](https://github.com/vega/vega-embed), meaning it is possible to pass a valid Vega spec as `spec` property and it renders the chart. It also offers the option to pass inline data as `dataValues` property (see docs).

It contains some default spec config, which is then deepmerged with the user-provided spec.

## Screenshots/Videos

image

- [x] All checks have passed


santilland:
This looks really nice! Thank you for preparing this, i like the possibility of being able to pass the data. 
@silvester-pari two things that popped into my head while checking the examples:
* what do we think about interactivity? So zoom and pan. This is not configured through the schema, or is it? Should we create an issue for this? In any case i don't think it is needed for the initial functionality.
* can we have some click event handler, to get information on what was clicked inside the chart? maybe also a separate issue?

silvester-pari:
> This looks really nice! Thank you for preparing this, i like the possibility of being able to pass the data. @silvester-pari two things that popped into my head while checking the examples:
> 
> * what do we think about interactivity? So zoom and pan. This is not configured through the schema, or is it? Should we create an issue for this? In any case i don't think it is needed for the initial functionality.
> * can we have some click event handler, to get information on what was clicked inside the chart? maybe also a separate issue?

There are lots of examples for chart interactivity at the [Vega-Lite website](https://vega.github.io/vega-lite/examples/#interactive). In the end, all of this is defined via the schema, but we can of course create a collection of preconfigured spec snippets/settings that are by default merged etc.
Regarding click handlers, there is a built-in way to add custom click handlers, so this should be doable too.

santilland:
Ok, so in theory anyone can define what interaction the want in the schema, thats good. So for sure we can later think if we should have some defaults at some point.
As for the event handling it would be great to have some helper callback property of something we can easily register to, but should also be doable at any point.
So I would say from my side, ok to merge, i will add the review.

santilland (approved):
Thanks for preparing this and the discussion on it, happy to try to start using it and evolving it together!

PR #1308, opened 2024-10-09
feat: line/area measurements

spectrachrome:

- [ ] All checks have passed

### DIFF
--- a/elements/map/src/draw.ts
+++ b/elements/map/src/draw.ts
@@ -3,6 +3,9 @@ import Draw, { createBox } from "ol/interaction/Draw";
 import { EOxMap } from "../main";
 import { Vector as VectorLayer } from "ol/layer";
 import { addNewFeature } from "../helpers";
+import { getArea, getLength } from "ol/sphere";
+import Overlay from "ol/Overlay";
+import Polygon from "ol/geom/Polygon";
 
 export type DrawOptions = Omit<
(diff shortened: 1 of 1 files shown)


elements/map/src/helpers/draw.js
```diff
@@ -77,3 +140,35 @@ export function addDraw(EOxMap, drawLayer, options) {
...
+  }
+  return output;
+};
+
+/**
+ * Format area output.
+ * @param {Polygon} polygon The polygon.
+ * @return {string} Formatted area.
+ */
+const _formatArea = function (polygon) {
```
silvester-pari, line 165: Is this used anywhere? It would be nice to support both lenght as well as area measurement.

silvester-pari (commented):
The way it is currently set up would mean we could add this rather as a functionality to `eox-map`, instead of `eox-drawtools`, right? Seeing that there is no logic in the drawtools itswelf (besides the overlay clearing)

elements/map/src/helpers/draw.js
spectrachrome: Yes, as currently implemented we would add it as a feature of the map, since that is where the OL interaction is created.

Currently this is not used, since the goal is mostly to determine distances in the GTIF apps. Question with two-dimensional measurements is also how the tooltip should behave (properly centered or stationary in a corner of the map).

PR #1317, opened 2024-10-16
Feat: Introduce Vector Layer feature selection

A-Behairi:
This pull request introduces a new feature for selecting vector layers features in the drawing tools and includes various enhancements and refactoring to support this functionality. 

- [x] All checks have passed


santilland:
Hi @A-Behairi , i think  for eodash it would be great if we can "autoenable" selection, do you think it would be possible to do that somehow that is not too complicated? or would it need to have a dedicated prop or something like that?
In the multi polygon selection, the list seems to still have some (at least) design issue. All features in the list seem to be highlighted (like they are selected). I mean i guess they are selected? So not sure what the behavior should really be. @silvester-pari any ideas?

silvester-pari:
I guess there are 2-3 possible "states".

For drawtools without VectorLayer:
- drawn (in list)
- selected (in list, highlighted)

For drawtools with VectorLayer:
- collection of pre-defined shapes (styling of original vector layer)
- "drawn" i.e. copied to the drawtools layer, in list but not highlighted
- selected (in list, highlighted)

Hover is also a selection btw., but could probably be styled differently than a click selection.

A-Behairi:
Thank you for the review, I also found another issue when switching from selection to drawing, that after the draw is completed the hover of the old selection layer  gets somehow activated again by the list, my guess is that it's related to map element updates but it needs further investigation. 

The "auto enable" of the selection wouldn't be too complicated to implement but it would mean that the drawtools element would listen to all of the map selection events all the time which I don't think is very ideal. Would it be fitting to eodash if added a programmatic interface to start/end/discard the drawing process instead ?

santilland:
> The "auto enable" of the selection wouldn't be too complicated to implement but it would mean that the drawtools element would listen to all of the map selection events all the time which I don't think is very ideal. Would it be fitting to eodash if added a programmatic interface to start/end/discard the drawing (or selection) process instead ?

Yes, i think we could start the listening when selecting an indicator which process has only a feature selection as input parameters (or something like that) and remove it as soon as another indicator is selected. That is a good idea. The question is can we do that from eodash or do we need to implement something here to be able to do that?

A-Behairi:
Yes, I think it will be best to implement a clear API on the element side. I will add a suggestion in this PR.

silvester-pari (changes requested):
Looking good! I'd just ask to change the story order, to keep the "Unstyled" story last.

santilland (approved):
Great! Thanks @A-Behairi functionality-wise i think its good to go. Some small test would be beneficial as Silvester pointed out ;)

silvester-pari (approved):

Issue #1251, opened 2024-09-12
Nested layers order


A-Behairi:
Nested layers inside group layers gets reversed sometimes. In my case this happened in [eodash](https://github.com/eodash/eodash), to investigate this further I debugged `oldLayers` & `newLayers` inside `set layers` property.

in the following image the map renders the layers in the right order
<img width="1800" alt="Screenshot 2024-09-11 at 13 09 50" src="https://github.com/user-attachments/assets/2cea6266-de2b-4c19-85bc-c5acdb30dbe6">

and the wrong order in this case:
<img width="1800" alt="Screenshot 2024-09-12 at 16 51 53" src="https://github.com/user-attachments/assets/d4de904e-c069-4787-8e69-0527f820b5a6">

although the layers passed have the same order

A-Behairi:
The layercontrol was removed in the explained example, I think it's more related to the sorting and reversing of the layers in the map.

silvester-pari:
Fixed in https://github.com/EOX-A/EOxElements/pull/1321

PR #1321, opened 2024-10-17
fix: layer order stability insided group layers

RobertOrthofer:
## Implemented changes

This PR fixes a problem described in #1251 about the stability of nested layers. The problem was a faulty conversion from the reverse-painter-order of the `EOxMap` layer json notation to the painter-order of `ol`.

The problem was quite hidden because it only affected the layer order inside group layers and only occured when `updating` the layers. A simple test has been added that sets the same layer json to a map, which actually reversed the layer order in `ol` until now.

- [ ] All checks have passed

### DIFF
--- a/elements/map/src/helpers/generate.js
+++ b/elements/map/src/helpers/generate.js
@@ -361,7 +361,7 @@ export function updateLayer(EOxMap, newLayerDefinition, existingLayer) {
 
     // Reorder the layers to match the REVERSE of the new definition
     // (EOx-Layers-JSON is in reverse painters order)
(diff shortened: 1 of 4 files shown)


silvester-pari (approved):
Thank you for this fix! Looks great, and yeah, we have reversed layers... Let's make this a topic for v2 :wink: 

I asked @A-Behairi to have a look if this fixes the issue he described, from my side it looks good to go.

A-Behairi (approved):
The issue doesn't occur anymore, Thank you for the fix!

PR #1334, opened 2024-10-28
fix: keep `selectLayer` activation state in change listener

silvester-pari:
## Implemented changes
This PR removes the hardcoded `setActive(true)` in the change listener for `selectLayer`. This caused issues if the selection was currently set to `active = false` and updating the layer would mistakenly reset the interaction back to `active`.

- [x] All checks have passed


RobertOrthofer (approved):
This looks good to me. This was added in https://github.com/EOX-A/EOxElements/pull/1078 to fix an issue with removing & adding a layer with a defined interaction, though setting it active here should not be needed

PR #1344, opened 2024-10-29
fix: initialize draw layer with `zIndex` 100

silvester-pari:
## Implemented changes

This initializes the draw layer with a `zIndex` of 100, so that it is on top of other layers in the majority of cases.

- [x] All checks have passed

### DIFF
--- a/elements/drawtools/src/main.js
+++ b/elements/drawtools/src/main.js
@@ -92,7 +92,7 @@ export class EOxDrawTools extends LitElement {
     this.draw = null;
 
     /**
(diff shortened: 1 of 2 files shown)


santilland:
I think this should work, @A-Behairi could you check integration in eodash, i must say i am somewhat surprised it is only one layer, the behavior in eodash seemed to support that there are multiple layers that hold the different feature types.
So a layer that keeps track of features for hover and one layer that keeps track of selected features.

A-Behairi:
I can confirm that the issue doesn't occur anymore in eodash. Thank you!

PR #1406, opened 2024-12-02
fix: view sync between a visible and a hidden view

silvester-pari:
## Implemented changes

This adds a small fix when syncing two views: for some reason, when syncing two views and the "receiving" one is not visible (and thus its view is not initially visible and usually defaults to 0 zoom if nothing else is defined), it "forces" its zoom onto the "giving" one (if no explicit zoom is set there as well). To mitigate this, when the "receiving" one is hidden **and** the "giving" one has no zoom set, it syncs the "other way round" in order for this zoom reset not to happen.

- [x] All checks have passed

### DIFF
--- a/elements/map/src/methods/map/setters.js
+++ b/elements/map/src/methods/map/setters.js
@@ -351,7 +351,17 @@ export function setSyncMethod(sync, EOxMap) {
       );
 
       // Set the view of the current map to match the view of the origin map
-      if (originMap) EOxMap.map.setView(originMap.map.getView());
+      if (originMap) {
+        // Check if the current map is hidden and the origin
+        // has no zoom - in this case the current map view
+        // takes over the zoom and resets it; so in this case it is
+        // better to sync the "other way round"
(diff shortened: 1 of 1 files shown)


RobertOrthofer:
The original behaviour is kind of expected, for `ol` technically there isn't a giving or receiving view, just 2 maps with the same view. Without a rendered container `ol` doesn't know how to calculate the `zoom` correctly.

I suppose you are hiding the map with `display: none;` (or not rendering the element at all)? If you just use `visibility: hidden;` it should work fine, if you want to keep a "master-slave" relation then this solution is fine as well

RobertOrthofer:
Did you manage to reproduce this in a vanilla ol setup? I can't quite reproduce this exact behavior...

Generally I suppose the current solution would be better than checking for the `visibility`-attribute, since the container could be 0x0 px for some reason, probably creating the same problem, but weird either way. Reversing the view-sync could potentially be dangerous because of other view-properties like `extent` or `rotation`

PR #1424, opened 2024-12-09
feat: result action

silvester-pari:
## Implemented changes
In this PR we add an event `click:result-action` that is emitted when an item's result action is clicked such that applications using `eox-itemfilter` can react when a user clicks on a result
TODO

- [x] PR description
- [x] Test
- [x] what if resultActionIcon is undefined
- [x] better name for property/event - maybe `click:result-action`?
- [x] support `card` display type
- [x] docs

## Screenshots/Videos

[Screencast from 2024-12-09 17:21:55.webm](https://github.com/user-attachments/assets/69b06a69-a927-408b-8a14-6f1b70fcba4b)

- [x] All checks have passed


silvester-pari:
I am questioning the usefulness of this. UX considerations: If the user sees an additional "action" button, (s)he might be tempted to always only click that, as it is not clear the list element can be clicked directly. Maybe there is an other way of achieving a similar result, or we need to first think about how to design this...

silvester-pari:
It doesn't seem a too uncommon design pattern, e.g.: https://vuetifyjs.com/en/components/lists/#action-with-text

image
image
image

silvester-pari:
I added some small improvements to the docs and to the card-rendering mode.

image

Thank you Radu, looks great now!

PR #1430, opened 2024-12-12
chore: replace bg colors with css variables

radupasparuga:
## Implemented changes
In this PR we remove some hard coded background colors and replace them with css variables so that elements are easier to customize

- [ ] All checks have passed

### DIFF
--- a/elements/itemfilter/src/style.eox.js
+++ b/elements/itemfilter/src/style.eox.js
@@ -21,7 +21,7 @@ export const styleEOX = `
     #000000 20%,
     transparent
   );
-  --background-color: #fff;
+  --background-color: var(--background-color);
   --padding: 0.5rem;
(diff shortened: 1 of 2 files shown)


silvester-pari (changes requested):
Thank you for looking into this! I had a quick look and found some small things:
- good idea to use the var, but maybe we should set it to transparent as fallback: `--background-color: var(--eox-background-color, transparent);`. Currently the fallback is white, and then we'd need to manually set it to transparent. Affects: drawtools (see editor story), itemfilter, layercontrol
- some elements are missing. Affects: itemfilter, jsonform (`.CodeMirror`)
 (see screenshots)

image
image
image
image
image

silvester-pari (approved):
:tada:

PR #1571, opened 2025-02-26
fix: initialize bboxes with value

silvester-pari:
## Implemented changes
This PR adds support for loading initial values for the `bounding-boxes`/`bounding-box` format.

## Screenshots/Videos

image

- [x] All checks have passed

### DIFF
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -1,14 +1,14 @@
 {
-  "elements/drawtools": "0.13.5",
+  "elements/drawtools": "0.14.0",
   "elements/itemfilter": "1.7.1",
-  "elements/layercontrol": "0.29.0",
+  "elements/layercontrol": "0.29.1",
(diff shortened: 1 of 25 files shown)


silvester-pari:
Thank you @srijitcoder - probably we need to add a check if `isBox()` is true? To only apply this to the bboxes format.
@A-Behairi maybe you could also have a quick look at this since you were the original author of the spatial editor.

silvester-pari:
@srijitcoder another thing is we should probably split up the PR and add the "parseOptions" and "animate" functionality separately, so we have different changelog entries for drawtools, map and jsonform...

A-Behairi (approved):
Looks good to me in general, one point that I notice that this will only work when creating the map inside the jsonform, although I'm not sure if we will have a use case that will need us to initialize values to an external map. Thank you for the PR!

PR #1601, opened 2025-03-11
fix: use dynamic imports to reduce bundle

### LINKED ISSUE #1588: Inclusion of ace editor creates huge set of bundle files
Since eox-jsonform has started supporting ace editor, the build output includes a large number of files related to the different languages ace editor supports. This is painful when trying to use the bundle of jsonform, or parent apps that use jsonform.

Explore ow this can be reduced, by i.e. only bundling used languages, and/or only include ace editor bundle if actually used.

Previously it was only two files:
- `eox-jsonform.js`
- `eox-jsonform.umd.cjs`

Now there are 478 files in total!

Another option would be to replace ace with monaco editor, but this would require writing a custom resolver for json-editor.

silvester-pari:
## Implemented changes

This PR makes use of the dynamic import functionality in ace editor (for modes, themes etc.) in order to reduce the amount of output bundle files. This brings the number of files in the dist folder from 478 back down to 2.

Fixes #1588.

- [x] All checks have passed

### DIFF
--- a/elements/jsonform/src/helpers/editor.js
+++ b/elements/jsonform/src/helpers/editor.js
@@ -121,25 +121,7 @@ export const createEditor = (element) => {
 
     // Check if any editor requires AceEditor
     const aceUsed = Object.values(editor.editors).find((e) => e?.ace_container);
-    // // console.log(editor.editors["root.code"].schema)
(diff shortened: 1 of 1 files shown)


A-Behairi (approved):
It builds as expected now 👍🏾, Thank you for the fix!

PR #1615, opened 2025-03-20
Fix: Refactor type `EoxLayer` & Generate Layer JSON Schema

A-Behairi:
## Implemented changes
This pull request aims to enhance type `EoxLayer` specifically and the types of the map element generally. Focusing on code improvements, type corrections, and the addition of the new script. The most important changes include adding `layers.ts` to include the new layer type definition and adding a new `schema:generate` script for schema generation from the layer type definition.

- [x] All checks have passed


silvester-pari (commented):
Thank you for this great update! From my end it looks good, but I would also like @RobertOrthofer to have a look since he has much deeper understanding of the eox-map code.

elements/map/src/controls/controls.js
RobertOrthofer, line 1: Maybe I'm blind, but it seems like `addInitialControls` is not used anywhere. Was this an old way to set basic controls via the parent application? Since it *seems* obsolete and is not documented, we could burninate it

RobertOrthofer:
The layer/source combinations seem fine. Some of them are deprecated but the superseding classes are included and I guess it's ok to support deprecated classes as long as they are still available in ol.

RobertOrthofer (approved):
From my side it also looks good. The map code has barely been touched and I approve the type changes with my limited knowledge about this topic.

PR #1633, opened 2025-03-26
fix: disable zoom in/out buttons when reaching limits

baloola:
## Implemented changes

This pr disables zoom in/out button when the max/min zoom level is reached.

## Screenshots/Videos

Screenshot from 2025-03-26 14-39-44

- [x] All checks have passed


elements/map/src/controls/controls.css
```diff
@@ -26,3 +26,9 @@
   width: 100%;
   height: 100%;
 }
+
+.disaibled-button {
+  opacity: 0.8;
+  filter: contrast(0.5);
+  cursor: not-allowed;
```
silvester-pari, line 33: It looks a bit strange with the reduced opacity and the button being so dark; how about we
- make it brighter
- disable pointer-events? 
so something like:
```
{
  filter: brightness(1.7);
  pointer-events: none;
}
```

silvester-pari (changes requested):

RobertOrthofer:
I definitely agree with the generic class name and the `pointer-events: none`.

My concerns are more general though: if I understand correctly, we calculate the `minZoom` and the `maxZoom` of the entire map via the highest maxZoom and the lowest minZoom of our layers that are explicitly set? But this only handles the zoom-control-buttons, the map can still be moved freely by user interaction or by programmatic moves. 

This also means that you should always set the minZoom and maxZoom explicitly. E.g. if you have a vector layer without minZoom/maxZoom and something like a "borders"-layer that you only want to display up to zoom level 10, the buttons will be disabled after zoom level 10 (I think we have a setup like this in eodash).

There are also the `minZoom` and the `maxZoom` properties of the view, which would be the correct place to restrict zooming, if desired. We are not utilizing this property yet, I only have early concerns that deriving minZoom and maxZoom from the layers could be some unexpected magic for some use cases.

elements/map/src/controls/controls.css
```diff
@@ -26,3 +26,9 @@
   width: 100%;
   height: 100%;
 }
+
+.disaibled-button {
```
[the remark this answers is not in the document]
baloola: reanamed the class into disabled

elements/map/src/controls/controls.css
```diff
@@ -27,8 +27,7 @@
   height: 100%;
 }
 
-.disaibled-button {
-  opacity: 0.8;
-  filter: contrast(0.5);
-  cursor: not-allowed;
+.disaibled {
```
silvester-pari, line 30: Please fix to "disabled" :wink:

RobertOrthofer (approved):

PR #1683, opened 2025-04-29
fix: check if parentElement exists before rendering

silvester-pari:
## Implemented changes

In Vue.js (and potentially other frameworks that rely on virtual DOM), initially `parentElement` is `null` for the `eox-layout-item`, causing the layout to break on first render (subsequent updates were working correctly).

This PR introduces a check if `parentElement` is defined, and if not, triggers the rendering only after one tick (after a `setTimeout`).

- [x] All checks have passed


A-Behairi (approved):
Thank you! this fixes it.

PR #1693, opened 2025-05-17
fix: add possibility to pass step to minmax custom filter

lubojr:
## Implemented changes

added passing of step to the minmax custom input for jsonform (only possitive integer input allowed in downstream library)

## Screenshots/Videos

image

- [x] For new features: I have updated existing story showcasing this new feature
- [ ] All checks have passed

### DIFF
--- a/elements/jsonform/src/custom-inputs/minmax.js
+++ b/elements/jsonform/src/custom-inputs/minmax.js
@@ -59,6 +59,8 @@ export class MinMaxEditor extends AbstractEditor {
     const attributes = {
       min: properties[minKey].minimum,
       max: properties[maxKey].maximum,
+      // only positive integer supported
+      step: properties[minKey].step || properties[maxKey].step,
       value1: startVals?.[minKey] || properties[minKey].default,
       value2: startVals?.[maxKey] || properties[maxKey].default,
       "slider-bg-fill": "#004170",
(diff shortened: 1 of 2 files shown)


lubojr:
@silvester-pari The failing pipeline on checking is not related to the PR content but seems to be caused by breaking changes in vega major version update done in https://github.com/EOX-A/EOxElements/pull/1646

silvester-pari:
Hm, that's strange, since in a local environment `tsc` runs fine. I'll see what I can do.

srijitcoder (approved):

silvester-pari:
I am trying my luck with https://github.com/EOX-A/EOxElements/pull/1694 but it is still failing to get the module declarations. Maybe @A-Behairi has an idea since he is the expert for this :thinking:

A-Behairi:
Vega and Vega light changed their TS module to [node16](https://github.com/vega/vega-lite/blob/a2853518ad34d4a7a9cbaa880aeedefeb1bcc803/tsconfig.json#L5), to support their TS now we have to change it as well, but this must ensure that all of dependencies we use export TS correctly based on the newer standards or they will fail. For example in eodash we didn't do this update because we have some packages that doesn't support node16.

silvester-pari:
Decided to revert vega: https://github.com/EOX-A/EOxElements/pull/1699

silvester-pari (approved):

PR #1694, opened 2025-05-19
chore(CI): update typecheck

silvester-pari:

- [ ] All checks have passed

### DIFF
--- a/elements/chart/src/enums/default-spec.js
+++ b/elements/chart/src/enums/default-spec.js
@@ -2,9 +2,7 @@
  * @type {import("vega-embed").VisualizationSpec}
  */
 export const DEFAULT_SPEC = {
-  // @ts-expect-error see https://vega.github.io/vega-lite/docs/size.html#specifying-responsive-width-and-height
   width: "container",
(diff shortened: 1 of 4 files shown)


silvester-pari:
> Vega and Vega light changed their TS module to [node16](https://github.com/vega/vega-lite/blob/a2853518ad34d4a7a9cbaa880aeedefeb1bcc803/tsconfig.json#L5), to support their TS now we have to change it as well, but this must ensure that all of dependencies we use export TS correctly based on the newer standards or they will fail. For example in eodash we didn't do this update because we have some packages that doesn't support node16.

@A-Behairi can we do this for a single element only? Since changing this globally causes 748 errors in 158 files :see_no_evil: 

Maybe by excluding the chart from the root tsconfig, and have a separate one there?

A-Behairi:
Yes I guess that would work. But the issue is that it would be a breaking change to `eox-chart` and all projects depending on it will have to change this and deal with a big amount of errors, some of which they may not have control over from other dependencies.

lubojr:
So taking a step back, did we really need to migrate the vega * packages to a newer major version including this?

silvester-pari:
Will try to revert for now: https://github.com/EOX-A/EOxElements/pull/1699

PR #1718, opened 2025-06-13
feat: coordinates, feature info (WMS) and data extraction (WebGL) for tooltips

santilland:
## Implemented changes

This PR adds extraction of map (layer) values during usage of the select interaction. It extends the automatic filling of map tooltips with data, which was previously limited to Vector (Tile) layers, adding coordinates (WebGLTile & WMS layers), feature info (WMS layers) and data values (WebGLTile layers) to the tooltip.

### Coordinates
- select interaction `options.coordinates: true` - shows the map coordinates for the current mouse position (works with click and hover interactions)
- select interaction `options.projection` - allows defining the projection (as object with properties `options.projection.name` and `options.projection.proj4_string`) for displaying the coordinates
- select interaction `options.precision` - allows defining the precision of the shown coordinates

### WMS Feature Info
- with a `options.condition: "click"` setup, it automatically fetches WMS feature info via `getFeatureInfo()` for `ImageWMS` and `TileWMS` layers and adds them to the tooltip (similar how GeoJSON feature properties are added to tooltips for vector layers). This only works for `click` interaction and not `pointermove`, as the `getFeatureInfo()` calls take some time

### Value Extraction 
- for `WebGLTile` type layers, the information of the layer provided by `getData()` for the current mouse position is added to the tooltip

The injected tooltip data can be rendered with the default `eox-map-tooltip`, or a custom tooltip.

## Screenshots/Videos

image

image

- [x] All checks have passed


elements/map/stories/coordinates-custom-tooltips.js
```diff
@@ -0,0 +1,148 @@
...
+          ],
+        },
+        interactions: [
+          {
+            type: "select",
+            options: {
+              id: "selectInteraction",
+              condition: "pointermove",
+              precision: 2,
+              coordinates: true,
```
silvester-pari, line 107: Could we show the coordinates in a separate example which is not WebGLTile? In order to make it more findable in the story list; ideally:
- one story for coordinates
- one story for WMS feature info
- one story for WebGLTile data extraction

elements/map/test/CustomTooltipCog.cy.js
silvester-pari, line 1: Could we rename this to something more aligned like e.g. `tooltipCogCustom.cy.js`? This might make it easier to find.

elements/map/test/CustomTooltipCog.cy.js
```diff
@@ -0,0 +1,14 @@
+import "../src/main";
+import "../src/plugins/advancedLayersAndSources/index";
+import { displayTooltipCogWithCustomTooltip } from "./cases/hover";
+
+/**
+ * Test suite for the EOX Map to load Tooltip when selecting a wms feature.
+ */
+describe("tooltip", () => {
+  /**
+   * Test case to display a WMS tooltip on a WMS layer
```
silvester-pari, line 10: Probably typo in comment (shouldn't be WMS?)

elements/map/test/fixtures/TooltipCogLayer.json
silvester-pari, line 1: In order to align with the other fixtures, please use camelCase, e.g. `tooltipCogLayer.json`

elements/map/stories/coordinates-custom-tooltips.js
baloola: since coordinates at the current implementation are provided with the selct layers ( vector/WebGL/WMS ). 
Probably having it within the WebGLTile layer story now would be okay. 
I'll create a separate issue for allowing coordinates for all othe types of layers.

elements/map/src/helpers/select.js
```diff
@@ -84,171 +89,345 @@ export class EOxSelectInteraction {
...
+              event.pixel[1] > this.eoxMap.offsetHeight / 2 ? "bottom" : "top";
+            overlay.setPositioning(`${yPosition}-${xPosition}`);
+            overlay.setPosition(event.coordinate);
+            let html = {};
+            //@ts-expect-error we do not expect type DataView
+            for (const [i, value] of data.entries()) {
+              const key = (i + 1).toString();
+              html[`band${key}`] = value; // Log each key-value pair in the data object
+            }
+            if (!hasNoZeros(html)) {
```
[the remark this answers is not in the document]
RobertOrthofer: I think the naming is correct, but I'm not positive if it does what we want. In the example the pixels outside the area of interest look like this:

```Js
{
    "band1": 0,
    "band2": 0,
    "band3": 0,
    "band4": 0,
    "band5": 0
}
```

So I suppose the `no data` pixels can be seen as `hasAllZeros`. Maybe there could be data pixels that have a real zero in only one of the bands, maybe because of a really dark spot or possibly a broken sensor. Pixels like that should be displayed I think. 
(`getData()` outside of the defined layer extent will return `null`, which is already handled)

silvester-pari (approved):

Issue #195, opened 2023-08-10
Dynamic feature clustering


silvester-pari:
Following [this OL example](https://openlayers.org/en/latest/examples/clusters-dynamic.html) and our implementation in [eodash](https://github.com/eurodatacube/eodash), it would be great to migrate this dynamic clustering as an option to the `eox-map`.
Ideally it should be able to cluster not only points, but any kind of features, if that makes sense; needs some brainstorming.

santilland:
I think we need to divide the expected logic into more specific steps:

start with:

- [ ] grouping of point features when close together (based on some logic)

afterwards:

- [ ] click interaction 
   - [ ] zoom in to cluster extent
   - [ ] if points overlap, open as "spider" selection to allow selecting one of the points

I would propose to start with the grouping concept and afterwards see how we can manage interaction.
There is an inherited cluster layer from vector layer, so we will need a dedicated cluster layer type. It would be best to see if all interactions etc work on that.

RobertOrthofer:
After some communication with @santilland , we detected a way to create the base-functionality of a clustered layer without the need of introducing new special Classes to the map element. Here is a basic `EOxLayer`:

```JSON
[
  {
    "type": "Vector",
    "properties": {
      "id": "clusterLayer"
    },
    "interactions": [
      {
        "type": "select",
        "options": {
          "id": "ClusterSelectInteraction",
          "condition": "click",
          "layer": {
            "type": "Vector",
            "properties": {
              "id": "clusterLayer"
            },
            "source": {
              "type": "Vector"
            },
            "style": {
              "stroke-color": "chartreuse",
              "stroke-width": 5
            }
          }
        }
      }
    ],
    "style": [{
        "filter": [">", ["get", "features", "length"], 1],
        "style": {
          "text-value": ["to-string", ["get", "features", "length"]],
          "circle-radius": 12,             
          "circle-fill-color": "#FFD700",  
          "circle-stroke-color": "rgba(255, 255, 255, 0.7)",
          "circle-stroke-width": 2,     
          "text-font": "normal 16px",         
          "text-fill-color": "#000",         
          "text-offset-y": 1
        }
      },
      {
        "else": true,
        "style": {
          "circle-radius": 12,             
          "circle-fill-color": "#FFD700",  
          "circle-stroke-color": "rgba(255, 255, 255, 0.7)",
          "circle-stroke-width": 2,     
          "text-font": "normal 16px",
          "text-value": "🗻",            
[...]

PR #1758, opened 2025-07-11
feat: cluster source and cluster-explode interaction

RobertOrthofer:
## Implemented changes

Implementation of a cluster a custom cluster explode / select interaction as defined in #195. This is an improved version of the cluster-implementation that has been used in `eodash`, made possible by creating a custom source `EOxCluster` which is available by enabling the `advancedLayersAndSources`. This new cluster source is based on the `OpenLayers` cluster source, but it comes with a custom `createCluster`-function, that allows clustering of polygon features, as well as a custom `geometryFunction`, that allows displaying points and polygons at the same time in a cluster layer, since the ol cluster source only displays point geometries by default. Additionally, if a cluster only consists of a single feature, the properties of that feature are written onto the cluster feature, allowing to style the cluster with a `FlatStyle` just like a normal vector layer.

Additionally, a custom `clusterExplode`-interaction has been implemented, which can be added to a vector layer with a cluster source, to enable zooming to the extent of clustered features as well as "exploding" the cluster into a spiderweb-shape so that the single features can be selected. If a single feature is clicked, a `CustomEvent` of the type `clusterSelect` is fired. The interaction itself is classified as "interaction" and not as "selectInteraction", but I am open to change that if needed.

A cluster is then exploded either if the `maxZoom` of the interaction is (almost) reached, or if the area of the extent of the clustered features is <= 100 (~ basically on top of each other).

Cursor handling is managed via the `handleEvent` of the interaction.

The style of the "exploded" features can be adapted by defining a `FlatStyle` as the `style`-property of the interaction. The style of the connecting LineStrings is hard coded, I suppose this too could be adapted.

A story and some tests have been added. The story points to [GeoJSON world Cities](https://raw.githubusercontent.com/drei01/geojson-world-cities/refs/heads/master/cities.geojson), a public GeoJSON containing polygons of cities.

As a side note: the layer generation ("`createLayer`") had to be slightly refactored, as a source of the type "`Cluster`" needs to be able to create their own source, which is a standard vector source.

## Screenshots/Videos
Screenshot of a cluster of polygons, single "unclustered" polygons at the sides
Screenshot 2025-07-14 at 15 35 58

- [x] All checks have passed


silvester-pari (changes requested):
Amazing addition! :tada: Thank you for the great work! It looks very promising and I'm very much looking forward to this.
I had a quick look at the demo story, and found two things that seem a bit odd:

1.) When disabling the `maxZoom` on the `clusterExplode` interaction, there still seems to be some limitation when clicking "deeper" - see the following video, where it's not possible to expand the features by clicking on the cluster; it expands to a certain zoom level, from there onwards I had to manually scroll further in order to show the two features; I would have expected that by clicking on the cluster, the two features would finally be revealed.

https://github.com/user-attachments/assets/c1c37fcc-1f6f-4942-8c6a-18c45f9294aa

2.) When clicking into a cluster (also `maxZoom` disabled), there are some situations where clicking on another cluster "zooms out" the view again; I believe it has to do with the actual features being at the exact same spot, one example is the following:

https://github.com/user-attachments/assets/fe7d2ad3-5a96-433f-b777-26d844d0baf9

RobertOrthofer:
@silvester-pari good points. The first issue was a logic problem - it is possible to be well below maxZoom (so no cluster explode happens), yet the features are not quite close enough together to explode. I have made a fallback case, if the logic would "zoom" to the clustered features, but no zoom actually happens (because they are too far apart), also explode the cluster. It feels a bit hacky but it seems to work fine - I am open to other ideas.

The second issue was a special case for polygons, I calculated the fit extent via the original geometry. I updated it to calculate the fit extent only with the `interiorPoint` of polygons, the same way we calculate it for the clustering itself.

RobertOrthofer:
I'd definitely avoid zooming out if possible in any way, because it could happen that a cluster could get bigger (number of clustered features increase) after clicking on a cluster. 

> Maybe, when clicking on a cluster and it can't split more, to force show the polygons?

Those polygons could be really small, and there might also be point features included in the cluster. Right now, if it can't split by zooming anymore, it explodes the cluster ([51d2a6d](https://github.com/EOX-A/EOxElements/pull/1758/commits/51d2a6dc2957181197b69817e21e297f235eba5c)). I'd have to wrap my head around how to detect and display large clusters so it would be useful to the user from a visual perspective without special cases making the features disappear. Either way, the remaining clusters / point features with same coordinates could end up overlaying the displayed polygons.

silvester-pari:
@RobertOrthofer maybe this is something we can look into in the future if such a usecase with overlapping polygons really should occur (probably won't). In any case, I'd propose to revert the change and not use the `interiorPoint` but the full geometry for the zoom extent, as zooming to the entire geometries looks much nicer, while zooming to the extent of the interior points feels a bit broken - that's my impression.

silvester-pari (approved):

PR #1828, opened 2025-09-17
fix: handle array values in update-url helper

danielfdsilva:
## Implemented changes

Ensure that array values (`{ key: [1,2,3]}`) in url parameters are encoded as `key=1&key=2&key=3` instead of `key=1,2,3`.

  - Didn't find any test file where this behaviour was tested. Please let me know where a test should be added.
- [x] All checks have passed


danielfdsilva:
I'm not sure what's the severity of this change. It is a pretty small fix, but if anyone is expecting the former array encoding, this will be a breaking change..

santilland (approved):
Looks good to me! At least in eodash we are not yet using this functionality. I don't think this is really used, but good point. Awesome, thank you @danielfdsilva !

PR #1903, opened 2025-10-29
fix(style): limit filter container height and make it scrollable

### LINKED ISSUE #1897: Itemfilter filters should be optionally scrollable
Currently only itemfilter results allow a scroll of results, but sometimes, we can have many expanded filters, that are cut and can not be scrolled. It would be great to either allow it or add a CSS variable allowing it - reference https://github.com/ESA-eodashboards/RACE-client/issues/19

silvester-pari:
## Implemented changes

This PR adds a maximum height for the filter container, and enables overflow/scrolling by default.
It tries to figure out how many filterProperties there are in order to set the minimum height and tries to find a compromise between filters and results display. Fixes #1897.

## Screenshots/Videos

image
image
image

- [x] All checks have passed

### DIFF
--- a/elements/itemfilter/src/main.js
+++ b/elements/itemfilter/src/main.js
@@ -480,7 +480,14 @@ export class EOxItemFilter extends LitElement {
         ${when(
           this.filterProperties,
           () => html`
-            <div style="display: var(--filter-display);">
+            <div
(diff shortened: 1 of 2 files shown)


lubojr (approved):
Thank you for a quick improvement! This is perfect solution.

PR #1915, opened 2025-11-03
fix: add getter for EOxMap zoomExtent

lubojr:
## Implemented changes

Fixes following warning from console:

`Field "zoomExtent" on EOxMap was declared as a reactive property but it does not have a getter. This will be an error in a future version of Lit. See https://lit.dev/msg/reactive-property-without-getter for more information.`

Although the getter does not make too much sense because the zoomExtent and does not correspond to current view extent and behaves internally as a method, this pleases lit, which would error out on next major version.

- [x] All checks have passed

### DIFF
--- a/elements/map/src/main.js
+++ b/elements/map/src/main.js
@@ -24,6 +24,7 @@ import {
   getLonLatCenterMethod,
   getLonLatExtentMethod,
   setZoomExtentMethod,
+  getZoomExtentMethod,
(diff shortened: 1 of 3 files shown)


lubojr:
Thank you for feedback. I added a getter. We already have a getter there for getting the projection in lat/lon, but this one for outputting in current map projection can be useful as well.

Issue #1905, opened 2025-11-03
Itemfilter aggregateResults empty value duplication within group


lubojr:
## Problem: 

Results are duplicated in a non-matching aggregatedResult key

## When

Issue happens if any theme in themes array for any of the items is changed to empty string when aggregateResults on themes is used.

## Reproduce

Can be observed in this storybook link - https://eox-a.github.io/EOxElements/?path=/story/elements-eox-itemfilter--card-display&args=items[0].themes[0]:

<img width="1586" height="1512" alt="Image" src="https://github.com/user-attachments/assets/481032a4-7668-4cc4-900a-0dc0e1e11f27" />

PR #1962, opened 2025-11-24
fix: aggregate falsy result values

silvester-pari:
## Implemented changes

This PR adds some measures to handle "falsy" data - previously empty strings or arrays (or missing properties) created two things that have been removed:
- "ghost categories", i.e. categories with an empty string name
- duplicated items, so items were added to the ghost category additionally to the actual aggregation slot

Note: this means that items with **only** falsy values (that don't match any aggregation slot) are **no more visible** in the results list; only aggregated results are shown. This is different than before, since beforehand the results that didn't fit into any aggregation slot would still be shown inside the "ghost" aggregation slot by mistake (albeit the others were duplicated). Comments and feedback welcome.

Fixes https://github.com/EOX-A/EOxElements/issues/1905.

## Screenshots/Videos

```json
[
  {
    "id": "foo",
    "title": "Foo Item",
    "themes": ["", "B"]
  },
  {
    "id": "bar",
    "title": "Bar Item",
    "themes": ["B", "C"]
  },
  {
    "id": "baz",
    "title": "Baz Item",
    "themes": ["A", "C"]
  }
]
```

### Before
image

### After
image

- [x] All checks have passed

### DIFF
--- a/elements/itemfilter/src/components/results.js
+++ b/elements/itemfilter/src/components/results.js
@@ -123,6 +123,21 @@ export class EOxItemFilterResults extends LitElement {
     toggleAccordion(event, this.config, this);
   }
 
+  #getAggregationProperties() {
+    const aggregationProperties = this.resultAggregation.filter(
+      (aggregationProperty) =>
+        this.aggregateResults(this.results, aggregationProperty).length,
+    );
+    const noCategoryResults = this.aggregateResults(
+      this.results,
+      "No category",
+    );
+    if (noCategoryResults.length > 0) {
+      aggregationProperties.push("No category");
+    }
+    return aggregationProperties;
+  }
+
   /**
    * Renders the HTML template for the component.
    */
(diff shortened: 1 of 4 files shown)


lubojr (changes requested):
Discussed internally that if we add the `No Category` fallback to not fully hide the entries, which have properties which do not match any group this is good to merge.
Thank you for the fixes!

silvester-pari:
I updated the logic, it needed some more changes as now the "No category" is created beforehand if there is an item that matches the criteria.

**Note that if an item has at least one valid category, it is shown there and not added to "No category" - only items are in "No category", if they don't fit anywhere else.** I think this would make the most sense, otherwise it would be strange that an item is inside one category, and then also in "no category", that would look like a bug. Opinions?

#### No valid categories
```js
      items: [
        {
          id: "foo",
          title: "Foo Item",
          themes: ["", ""],
        },
        {
          id: "bar",
          title: "Bar Item",
          themes: ["B", "C"],
        },
        {
          id: "baz",
          title: "Baz Item",
          themes: ["A", "C"],
        },
      ],
```

image


#### One matching category
```js
      items: [
        {
          id: "foo",
          title: "Foo Item",
          themes: ["", "B"],
        },
        {
          id: "bar",
          title: "Bar Item",
          themes: ["B", "C"],
        },
        {
          id: "baz",
          title: "Baz Item",
          themes: ["A", "C"],
        },
      ],
```

image

lubojr (approved):
Yes, this is perfect addition! Thank you for considering this other case with falsey and valid value.

PR #1984, opened 2025-12-04
fix: Support partial layer ID matching for selection

A-Behairi:
## Implemented changes
This updates the logic for retrieving a layer by its ID in the `initSelection` method. Instead of using the `getLayerById` method, it now searches through for a layer whose ID starts with the given `layerId`.

- [ ] All checks have passed

### DIFF
--- a/elements/drawtools/test/_mockMap.js
+++ b/elements/drawtools/test/_mockMap.js
@@ -37,6 +37,7 @@ export class MockMap extends HTMLElement {
           },
         ],
       }),
+      getAllLayers(){ return this.getLayers().getArray()},
       // Simulating removeLayer method
       removeLayer() {},
     };


silvester-pari:
Hi, thank you, but I am unsure how exactly this would behave in some situations. Let's say one has a layer `selectLayer` and a layer `selectLayer2` - looking for `selectLayer` now could match either of those, depending on the rendering order within the map. I'd consider this a breaking change 🤔 

~Also, the `getLayerById()` helper function recursively finds and gets layers, also within groups within groups, whereas with this PR it's limited to the topmost layer collection (not working inside groups as it only does `getLayers().getArray()`. Also that's a breaking change imo.~ Seeing that the ol `getAllLayers()` is supposed to get all layers from groups. Haven't used that yet, but if it's doing the same thing then that should be fine.

Are my concerns valid, or am I misunderstanding?

lubojr:
Good point, custom wildcard support would be a better solution.

santilland:
The wildcard idea does not sound bad, but that would mean we would need to again educate the user, that in the jsonform they need to add a star (*) to the collection identifier in order to match.
Maybe somehow eodash can always inject a star before passing it to the json-form?

santilland:
How do we differentiate regex (wildcard) vs exact match in getLayerById?
Should we maybe introduce a new function, getLayerByRegex?

PR #1977, opened 2025-12-01
feat!: reverse layer order

silvester-pari:
## Implemented changes

This PR introduces a breaking change: the ordering of the layers inside the `layers` array has been reversed; this has been done in order to conform with all other mapping libraries (meaning the further one progresses in the array, the "higher" the layer sits on the map).
For eox-map v1, a reversed layers array was a decided back then in order to "make it more intuitive" for developers, meaning layers in the array within the code editor represented the visualization on the map ("higher" up in the array in the code, "higher" up in the map). This turned out to be the opposite of intuitive, and now the decision has been made to un-reverse the layer order again.
This change affects all apps using the `eox-map`, as well as other components which use the map (such as `eox-layercontrol`).

- stories and tests updates: #1985

## Checklist before requesting a review

- [x] I have performed a self-review of my code
- [x] I have added a test related to this feature/fix
- [x] For new features: I have created a story showcasing this new feature
- [x] All checks have passed
- [x] The PR title [follows conventional commit style and reflects the type of change (major/minor/patch, breaking)](https://github.com/googleapis/release-please?tab=readme-ov-file#how-should-i-write-my-commits)
- [x] The PR title describes the change within this element (each merged PR equals one entry in the element's changelog!)

PR #1988, opened 2025-12-07
feat: serialize layers allowing objects, functions, dates etc.

silvester-pari:
## Implemented changes

This PR replaces `JSON.stringify()` with serialize from the [serialize-javascript](https://github.com/yahoo/serialize-javascript) package. This add support for objects, functions, dates etc. inside the layers array, which previously broke the `JSON.stringify()`, thus not allowing e.g. `tileLoadFunctions` etc.

Depends on #1977.

This is now possible:
```js
    layers: [
      {
        type: "Tile",
        source: {
          type: "OSM",
          tileLoadFunction: (imageTile, src) => {
            console.log(src);
            imageTile.getImage().src = src;
          },
        },
      },
    ],
```

Note that passing DOM objects are still not possible, as they are not supported by `serialize-javascript`.

- [x] All checks have passed

### DIFF
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -1,6 +1,6 @@
 {
   "elements/drawtools": "1.1.5",
-  "elements/itemfilter": "1.13.0",
+  "elements/itemfilter": "1.13.1",
   "elements/layercontrol": "1.3.1",
   "elements/map": "1.26.3",
   "elements/stacinfo": "1.0.2",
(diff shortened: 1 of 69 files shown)


RobertOrthofer:
Ah, now I see it, thanks for the clarification. I agree that "same function with different whitespace" is a non-issue here. I think this is good, we somehow need to document the feats of allowed functions (e.g. "self-contained" and similar)

PR #2145, opened 2026-02-16
feat: 3d globe LOD option

baloola:
## Implemented changes

This PR makes the Globe resolution reduced, the behaviour would be that by default the 3D globe would use a bigger grid size, which will make the rendered scene appear as if the camera is on a lower zoom level ( bigger height). This behavior helps immensely in the performance.
The behavior can be reset to the default value by setting the property `globeOptions.useHighLOD` to true.
```json
globeOptions.useHighLOD: true
```

## Screenshots/Videos
currnt default resolution (level of detail):

Screenshot from 2026-02-19 20-27-32

setting `useHighLOD` to true: 
 
Screenshot from 2026-02-19 20-15-16

- [x] All checks have passed

### DIFF
--- a/elements/map/src/main.js
+++ b/elements/map/src/main.js
@@ -246,7 +246,7 @@ export class EOxMap extends LitElement {
   last2dProjection = "EPSG:3857";
 
   /**
-   * Whether use less level of details or not.
+   * Whether use high level of details or not.
    */
   #useHighLOD = false;
 
@@ -516,7 +516,7 @@ export class EOxMap extends LitElement {
   }
 
   /**
-   * Sets if the LOD level for the globe will not be reduced. This can be used to increase the level of detail for the globe, but may have performance implications.
+   * Sets if the (default) higher LOD level for the globe will not be used. This can be used to increase the level of detail for the globe, but may have performance implications.
    *
    * @param {boolean} useHighLOD - The useHighLOD.
    */


elements/map/src/main.js
```diff
@@ -157,6 +157,7 @@ export class EOxMap extends LitElement {
       selectInteractions: { attribute: false, state: true, type: Object },
       sync: { attribute: "sync", type: String },
       zoomExtent: { attribute: false, type: Array },
+      useDefaultLOD: { attribute: false, type: Boolean },
```
silvester-pari: This can be an attribute, e.g. called `use-default-lod`. This way one could use `<eox-map projection="globe" use-default-lod></eox-map>` in the HTML template without any need of JS.

elements/map/src/main.js
```diff
@@ -509,6 +515,23 @@ export class EOxMap extends LitElement {
...
+   * Sets if the LOD level for the globe will not be reduced.
+   *
+   * @param {boolean} useDefaultLOD - The useDefaultLOD.
+   */
+  set useDefaultLOD(useDefaultLOD) {
+    this.#useDefaultLOD = useDefaultLOD;
+  }
+
+  /**
+   * Gets the current useDefaultLOD state of the map.
```
silvester-pari: Please add some better description here what the property does, for someone completely new to eox-map. The description here is rendered in the element's docs args table, so it is the most important!

elements/map/src/main.js
baloola: added

PR #2215, opened 2026-03-30
fix: update 3D (XYZ) layers url's when params changed in layerConfig

baloola:
## Implemented changes
This PR updates the XYZ layer based on the changes in the layerConfig in 3D 

Originally the changes in the layer config are not reflected in 3D, this PR allows the 3D scene to adapte the changes in the layerControl element.

## Screenshots/Videos
[Screencast from 2026-03-30 17-23-01.webm](https://github.com/user-attachments/assets/48cbfae1-a7c3-48e1-882b-2a4dd1a4129a)

- [x] All checks have passed


silvester-pari (changes requested):
Thank you for the PR. However, this is mixing quite some things between layercontrol and map/globe. I am wondering if this could be split up better, e.g. adding a listener in eox-map if the source triggers maybe a change or propertychange (not sure how to listen to source url change).

In any case, it would be great to split this PR up into two parts, the layercontrol part and the map part, as each PR equals to one changelog entry. If on both sides it's a fix and the commit/CHANGELOG message can be the same then it's fine I guess, but usually we split this up, as usually on one side it's a fix and on the other a feature or vice versa. 🤔

silvester-pari (approved):
Thank you for splitting up the PRs! Even though the current setup is not ideal (using the globe API from inside the layercontrol), I also don't have a better idea at the moment; let's merge for now so we can include the bug fix, if we run into issues down the line we can revisit this.

PR #2228, opened 2026-04-04
feat(drawtools): add programmatic control APIs

alek-cesarz:
## Summary

Adds public APIs for programmatic control of drawtools, enabling applications to manage draw state and features without reaching into internal properties.

- **`stopDrawing()`** — deactivates draw interaction without discarding features (complements existing `startDrawing()` / `discardDrawing()`)
- **`suppressEvents`** property — when `true`, prevents `drawupdate` event emission during programmatic feature changes (avoids feedback loops)
- **`removeFeatureByIndex(index)`** — removes a specific feature from the draw layer by index; returns `boolean`
- **Escape key fix** — prevents listener accumulation on repeated start/stop cycles; delegates to `stopDrawing()` for consistent state cleanup

Also includes the modify interaction key fix from #2227 (can be rebased once that merges).

## Motivation

Applications using drawtools for multi-feature editing (e.g., multiple AOIs) need to:
1. Stop drawing without losing existing features — previously required `draw.setActive(false)` + `currentlyDrawing = false` + `drawLayer.set('isDrawingEnabled', false)` (internal properties)
2. Programmatically remove individual features — previously required accessing `drawLayer.getSource().removeFeature()` (internal OL source)
3. Suppress events during bulk operations — previously required external guard flags since `drawupdate` fires on any source change

These APIs make these operations first-class without touching internals.

## Changes

| File | Change |
|------|--------|
| `src/methods/draw/stop-drawing.js` | New: `stopDrawingMethod` |
| `src/methods/draw/start-drawing.js` | Fix: remove old Escape handler before adding new; delegate to `stopDrawingMethod` |
| `src/methods/draw/emit-drawn-features.js` | Gate `drawUpdateEvent()` with `suppressEvents` check |
| `src/methods/draw/index.js` | Export `stopDrawingMethod` |
| `src/methods/draw/init-layer.js` | Fix: modify interaction key (`drawInteraction_modify`) |
| `src/main.js` | Add `suppressEvents` property, `stopDrawing()`, `removeFeatureByIndex()` methods |
| `test/_mockMap.js` | Fix: `removeFeature` respects feature argument |
| `test/cases/stop-drawing.js` | New test |
| `test/cases/suppress-events.js` | New test |
| `test/cases/remove-feature-by-index.js` | New test |
| `test/cases/index.js` | Export new tests |
| `test/general.cy.js` | Register new tests |
| `README.md` | API documentation with methods/properties/events tables |

## Test plan

- [ ] All existing Cypress tests pass
- [ ] New test: `stopDrawing()` deactivates draw, preserves features
- [ ] New test: `suppressEvents = true` prevents `drawupdate`, `false` allows it
- [ ] New test: `removeFeatureByIndex(0)` removes correct feature, out-of-bounds returns `false`
- [ ] Manual: draw features, call `stopDrawing()`, verify features remain on map

### DIFF
--- a/elements/drawtools/README.md
+++ b/elements/drawtools/README.md
@@ -16,7 +16,60 @@ import "@eox/drawtools/dist/eox-drawtools.js"
 <eox-drawtools></eox-drawtools>
 ```
 
-Please refer to the [drawtools docs](https://eox-a.github.io/EOxElements/?path=/docs/elements-eox-drawtools--docs) for more details.
+## API
+
+### Methods
+
+| Method | Description |
+|--------|-------------|
+| `startDrawing()` | Activates the draw interaction on the map |
+| `stopDrawing()` | Deactivates the draw interaction without discarding existing features |
+| `discardDrawing()` | Stops drawing and clears all drawn features |
+| `removeFeatureByIndex(index)` | Removes a feature at the given index. Returns `true` if removed, `false` if out of bounds |
+| `emitDrawnFeatures()` | Manually triggers feature emission and `drawupdate` event |
+
+### Properties
+
+| Property | Type | Default | Description |
+|----------|------|---------|-------------|
+| `type` | `String` | `"Polygon"` | Draw type: `Polygon`, `Box`, `Point`, `Circle`, `LineString` |
+| `multiple-features` | `Boolean` | `false` | Allow drawing more than one feature |
+| `allow-modify` | `Boolean` | `false` | Allow modifying features after drawing |
+| `continuous` | `Boolean` | `false` | Auto-start next drawing after completing one |
+| `show-list` | `Boolean` | `false` | Display a list of drawn features |
+| `show-editor` | `Boolean` | `false` | Show GeoJSON editor |
+| `import-features` | `Boolean` | `false` | Allow drag-drop and file upload import |
+| `format` | `String` | `"feature"` | Output format: `feature`, `geojson`, `wkt` |
+| `projection` | `String` | `"EPSG:4326"` | Projection of emitted features |
+| `suppressEvents` | `Boolean` | `false` | When `true`, suppresses `drawupdate` event emission. Set before programmatic feature changes to avoid feedback loops |
+
+### Events
+
+| Event | Detail | Description |
+|-------|--------|-------------|
+| `drawupdate` | `Feature[] \| GeoJSON \| WKT` | Fires when features are added, modified, or discarded (unless `suppressEvents` is `true`) |
+
+### Programmatic Control Example
+
+```javascript
+const drawtools = document.querySelector("eox-drawtools");
+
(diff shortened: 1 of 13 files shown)


silvester-pari (commented):
Thank you very much @alek-cesarz for your contribution! Here are my comments to the things you implemented:
- `stopDrawing()`: what would be a use case for this? Currently we have interfaces where a single feature can be drawn (then it stpos automatically) or multiple are allowed; we didn't encounter a situation where the drawtools need to "stop", so I would be curious to understand. In general, I am unsure if a stopDrawing() API makes sense if there is no corresponding UI button (like exists for `start` and `discard`
- `suppressEvents`: this is a good idea to fix this, but I wonder if this shouldn't be the default? I don't see a use case where the developer would **not** want to suppress feedback loops - maybe I am missing something?
- `removeFeatureByIndex(index)`: I can see this as a good feature, both by index and actually also maybe a `removeFeature(feature)`
- escape key fix: Thank you for this fix, sounds very useful

In general, we usually have one fix/feature per PR in this repository, since each PR equals to one changelog entry.  It makes tracking changes much easier and is recommended for this repo unless the fixes/features are really closely intertwined. Therefore I'd like to ask you to split up the contributions into separate PRs, e.g. starting with the escape key fix and the event loop suppression fix, and the removeFeatureByIndex/removeFeature feature. Regarding the stopDrawing it depends on the outcome of our discussion.

alek-cesarz:
@silvester-pari, 
in my case I needed a way to control the drawing mode both via Esc and a custom toggle icon. this allows any event to switch it off - exactly as you have written. adding a corresponding button to drawtools makes it complete.
agreed, making it default covers most cases.

silvester-pari:
Hello again @alek-cesarz, here's an update on your proposals made in this PR:
- `drawUpdateEvent` loop: addressed and fixed in https://github.com/EOX-A/EOxElements/pull/2237
- escape key event listener accumulation: addressed and fixed in https://github.com/EOX-A/EOxElements/pull/2300
- `removeFeatureByIndex(index)` added in https://github.com/EOX-A/EOxElements/pull/2302

All of these changes are already available in `eox-drawtools` `v1.4.0`! :tada: 

The only thing not addressed yet is the `stopDrawing` functionality, as I am not sure if I have yet understood the use case. Could you describe the UX flow in your application where a "stop drawing" situation occurs?

silvester-pari:
@alek-cesarz after discussing this with the team, we came to the conclusion that offering a `stopDrawing` method indeed makes sense as it would just be the programmatic equivalent of pressing the `Esc` key. It has now been implemented in https://github.com/EOX-A/EOxElements/pull/2333.

Sorry it took so long and thank you for your patience! I think now all of the things you proposed are implemented, therefore I am closing this PR. Feel free to comment or reopen at any time.

PR #2238, opened 2026-04-08
fix: isolate feedback tsconfig, update type generation to include jsonform types

lubojr:
## Description

Updates feedback tsconfig to allow running types:generate standalone with types being in place where package.json announces it.

## Warning
new helper step `"pretypes:generate": "cd ../jsonform && npm run types:generate"` is not platform agnostic

I can remove this step if that is a problem (user is then supposed to run npm run types:generate in jsonform folder on their own, as was up to now)

- [x] All checks have passed


silvester-pari (commented):
Thank you for this. I believe that we have this situation in some other elements as well (one element including/being dependent on another element), until now I thought it worked since we ran it for all elements from the repo root. I'd suggest if we do this for eox-elements, then it would be good to also do it for the other elements when there are inter-element dependencies. Although I am not sure I understand why eox-feedback would need to be built separately and not within a pipeline?

Regarding `rootDir` and `paths` - if I understand correctly, this is definitely something we should add to the other elements as well?

lubojr:
@silvester-pari eox-feedback does not need to have it built individually, it does work overall that the global types are built, but when I wanted to conditionally import the eox/feedback like so:
```
if (!customElements.get("eox-feedback")) {
  await import ("@eox/feedback");
}
```
(I need to conditionally import it like that because in vitepress pages deployment, we already import it once in the template initialization and second - later import in eodash throws errors because the eox-feedback always defines the same web component name).

then without the PR I've created it throws a tsc error
`error TS2307: Cannot find module '@eox/feedback' or its corresponding type declarations.`

Regarding rootDir and paths could be added to all other elements, yes, I can have a look at adding these.

For now I am adding `//@ts-expect-error` in eodash.

PR #2243, opened 2026-04-10
feat: timeline clustering

### LINKED ISSUE #2197: Review usage of bins for timeline
Currently bins are kind of used for the timeline in an not completely defined way. There is a default 1 day binning, which fits also to the datepicker selection.
This is in principle good, but it would be great to better define the expected bin utilization.

So i think questions to answer are:
1) should we use binning at all?
   We could not bin anything, then datetime items would need to have a "minimal" size.
2) if we want to bin, is a fixed binning good? i.e. always daily binning to match datepicker selection/view
3) if not could we introduce a dynamic binning?, i.e. binning size changes when zooming in or out

My proposal would be to 1) yes, use binning, and 2) keep it fixed to 1 day.

This would mean:
* datetime entries are always inside of 1 day
* default selection range size can be kept to the day range (so clicking randomly on an area of the timeline does the same action - select a day time range - as clicking on a day in the datepicker mode)
* ideally we can use the [clustering function](https://visjs.github.io/vis-timeline/examples/timeline/other/clustering.html) from the vis-timeline ?

srijitcoder:
## Implemented changes
- Introduced clusters to the timeline. While zooming in, all the clusters will be split into dots, and while zooming out, all the dots will be clustered to form bins 
- We can enable or disable clusters using `clustering="true | false"` attribute. By default it is disabled.
- Clusters cannot be clicked. If we double click it will zoom in to dots.
- Dots can be clicked, and a single dot will be selected, and on hover, if there is more than one item, it will show the item tooltip. 
- Tested it with 3.7K point data sent by @lubojr  and it is working well. 

```html
<eox-timecontrol>
  <eox-timecontrol-timeline .clustering=${true}></<eox-timecontrol-timeline>
<eox-timecontrol>
```

## Screenshots/Videos

https://github.com/user-attachments/assets/80096478-4202-4781-8197-ca3a39f61a38

- [x] All checks have passed


lubojr:
Example from timeslider in old eoxc with ~4k points: https://vs.demo.hub.eox.at/?x=12.000000&y=49.000000&start=2023-06-01T00%3A00%3A00Z&end=2023-06-01T23%3A59%3A59Z&z=6

Extracted S2 test data with 3700 items:
 [opensearch_results.json](https://github.com/user-attachments/files/27432011/opensearch_results.json)

elements/timecontrol/src/methods/timecontrol/first-updated.js
```diff
@@ -177,10 +177,10 @@ export default function firstUpdatedMethod(EOxTimeControl, emitUpdateEvent) {
                 EOxTimeControlTimeline.visTimeline.setOptions({
                   ...EOxTimeControlTimeline.visTimeline.setOptions,
                   start: dayjs(dateRange[0])
-                    .subtract(30, "day")
+                    .subtract(10, "day")
```
silvester-pari: What is the rationale behind changing this from one month +/- to 10 days +/-?

elements/timecontrol/src/methods/timeline/init-timeline.js
```diff
@@ -269,6 +352,23 @@ export default function initTimelineMethod(EOxTimeControlTimeline) {
...
+          const isClusterItem = dom
+            .querySelector(`.item-${originalItemData.id}`)
+            .classList.contains(CLUSTER_ITEM_CLASSNAME);
+          const filteredItems = filter(items, {
+            originalDate: originalItemData.originalDate,
+            group: originalItemData.group,
+          });
+          return isClusterItem || filteredItems.length === 1
+            ? false
+            : `<span>${filteredItems.length} item${filteredItems.length > 1 ? "s" : ""}</span>`;
```
silvester-pari: Could we align this more to match the other tooltip that we have on the date picker (including dark mode support by using CSS variables)?

Image


Image


Image

elements/timecontrol/src/styles/style.timeline.js
```diff
@@ -61,21 +61,33 @@ export const styleTimeline = `
...
-.vis-item.milestone.vis-point:hover {
+.vis-item.milestone.vis-point:hover:not(.vis-cluster-item) {
   background: #ffc200 !important;
   cursor: pointer;
 }
+.vis-item.milestone.vis-point.vis-cluster-item {
+  border-radius: 0px !important;
+}
+.vis-item.milestone.vis-point.vis-cluster-item-start {
+  border-top-left-radius: 6px !important;
```
silvester-pari: The border radius does not seem to apply in some cases: 

Image

Is this intended or still to be fixed?

elements/timecontrol/src/methods/timeline/init-timeline.js
```diff
@@ -269,6 +352,23 @@ export default function initTimelineMethod(EOxTimeControlTimeline) {
...
+            .dom.root;
+          const isClusterItem = dom
+            .querySelector(`.item-${originalItemData.id}`)
+            .classList.contains(CLUSTER_ITEM_CLASSNAME);
+          const filteredItems = filter(items, {
+            originalDate: originalItemData.originalDate,
+            group: originalItemData.group,
+          });
+          return isClusterItem || filteredItems.length === 1
+            ? false
```
silvester-pari: Could we also show the tooltip on the clusters, showing the total number of items inside the cluster? This would be useful so the user knows how big a cluster is.

silvester-pari (changes requested):
Thank you! It already looks great, however I have a few questions/suggestions.

silvester-pari (approved):

PR #2245, opened 2026-04-13
feat: configuration for timeline selection defaults (duration, resize)

### LINKED ISSUE #2202: Additional configuration options
- default range: currently the default range on click is set to one day: https://github.com/EOX-A/EOxElements/blob/main/elements/timecontrol/src/methods/timeline/init-timeline.js#L172-L173. This could be customizable to a different time range (e.g. one week, one hour, one second, etc., maybe also allowing "0 seconds"). Required to allow setting e.g. a single point in time (0 seconds) instead of a range
- range modification: currently, the user is always allowed to modify the selection range. In cases where we have e.g. "0 seconds" (see above), we might prevent the user to modify the range. Therefore we could add a configuration option to allow/disallow range modification

These two options make the configuration quite flexible; on the other hand, we could instead introduce a "single point in time" mode which sets both properties (so 0 seconds, not allowed to modify range). Not sure if that would be enough or if we need the granular settings.

srijitcoder:
## Implemented changes
Added two configurations to control range selection in `eox-timecontrol-timeline`:
- `selectionDuration`: Defines the default selection duration when clicking on the timeline background. It strictly supports native [dayjs duration](https://day.js.org/docs/en/durations/creating) formats, including ISO 8601 strings (e.g., `"P1M"`, `"PT0S"`), duration objects (e.g., `{ months: 1 }`), or a number (milliseconds).
- `selectionResizable`: Controls whether the selected range can be resized or dragged by the user. If set to `false`, the user cannot modify the range manually.

Two example stories have been added:
1. **TimelineSelectionDuration**: Demonstrates setting default selection ranges.
2. **TimelineSelectionResizable**: Demonstrates locking the selection range.

```html
<!-- Example: One month fixed duration -->
<eox-timecontrol for="eox-map#timeline">
  <eox-timecontrol-timeline
    .selectionDuration=${{ months: 1 }}
    .selectionResizable=${false}
   ></eox-timecontrol-timeline>
</eox-timecontrol>

<!-- Example: Point selection (0 seconds) using ISO 8601 string -->
<eox-timecontrol for="eox-map#timeline">
  <eox-timecontrol-timeline
    .selectionDuration=${"PT0S"}
   ></eox-timecontrol-timeline>
</eox-timecontrol>
```

- [x] All checks have passed


elements/timecontrol/src/components/timecontrol-timeline.js
```diff
@@ -30,6 +31,8 @@ export class EOxTimeControlTimeline extends LitElement {
   static get properties() {
     return {
       unstyled: { type: Boolean, attribute: "unstyled" },
+      selectRangeType: { type: String, attribute: "select-range-type" },
```
silvester-pari: I am unsure about the naming - "range type" sounds like something else, maybe it can be `selectionDuration` or `defaultDuration`?

elements/timecontrol/src/components/timecontrol-timeline.js
```diff
@@ -30,6 +31,8 @@ export class EOxTimeControlTimeline extends LitElement {
...
+      rangeSelection: { type: Boolean, attribute: "range-selection" },
```
silvester-pari: Same here, maybe call it `selectionResizable`? So we would have e.g. `selectionDuration` and `selectionResizable`?

elements/timecontrol/src/components/timecontrol-timeline.js
```diff
@@ -59,6 +62,21 @@ export class EOxTimeControlTimeline extends LitElement {
...
      */
     this.unstyled = false;
+
+    /**
+     * The selected range of the timeline.
+     * Use any unit type from dayjs - https://day.js.org/docs/en/manipulate/start-of#list-of-all-available-units
+     *
+     * @type {selectRangeType}
+     */
+    this.selectRangeType = "day";
```
silvester-pari: Here it probably would be better to use **durations**? https://day.js.org/docs/en/plugin/duration

If I understand correctly, this allows numbers as well as [ISO 8601 durations](https://en.wikipedia.org/wiki/ISO_8601#Durations).

Maybe it then also supports "0" as duration (`PT0S` in ISO 8601).

elements/timecontrol/src/methods/timeline/init-timeline.js
```diff
@@ -303,15 +311,20 @@ export default function initTimelineMethod(EOxTimeControlTimeline) {
       updateRangeElements(EOxTimeControlTimeline);
       handleTimelineChanged(EOxTimeControlTimeline, EOxTimeControl, options);
     });
-    visTimeline.on("timechange", () =>
-      updateRangeElements(EOxTimeControlTimeline),
-    );
+    visTimeline.on("timechange", () => {
+      if (EOxTimeControlTimeline.rangeSelection)
+        updateRangeElements(EOxTimeControlTimeline);
```
silvester-pari, line 332: Does this (or `EOxTimeControlTimeline.setDateRange`) emit an event that a developer could listen to?

silvester-pari (changes requested):

elements/timecontrol/src/methods/timeline/init-timeline.js
srijitcoder: Yeah, both of them are emitting an event called `select`

silvester-pari (approved):

PR #2304, opened 2026-04-30
feat: support stringified JSON for attributes

silvester-pari:
## Implemented changes

This PR adds support for stringified JSON to all elements attributes. 

### Reasoning
The reason for this is that the difference between HTML attributes and JS properties is blurred by many popular JS frameworks and many peolpe (as well as LLMs) are not aware of the difference. Especially when working with LLMs, it is a constant "education" effort as they assume everything can be passed as stringified attribute.

### Changes for existing and future code using EOxElements
What this means going forward: no breaking change, but hopefully easier onboarding of new developers and better support for LLMs.

### Pros/Cons
One downside of this is the shift of "educating" devs/LLMs of when to switch to JS properties instead of HTML attributes (e.g. if the object includes complex things like date objects, functions, etc., then it is **still required** to use JS properties), but these edge cases should be much less common than the current issues.

Another downside is moving away from a "clean" separation of concerns according to web component standards, but hopefully with enough benefits (as described above) to justify this move.

### EOxElements example stories
Together with this change, the story code snippet rendering has been adapted to the following logic:
- if a property is not stringifiable (e.g. including functions, date objects, etc.), still render a JS property
- if a property is stringifiable and the string length is longer than 50 characters, still render a JS property (in order to avoid bloated HTML templates)
- if a property is stringifiable and 50 characters or less, then render it as an attribute

### Versioning
This PR will bump all the affected elements in their minor version.

## Screenshots/Videos

image

- [x] All checks have passed

### DIFF
--- a/.storybook/custom-panels/helpers/parse-elements.js
+++ b/.storybook/custom-panels/helpers/parse-elements.js
@@ -24,11 +24,17 @@ export const parseElements = (storyData) => {
     const storySlot = !!elementArgs.storySlot;
     const storyWrap = !!elementArgs.storyWrap;
 
-    const attributes = Object.entries(elementArgs).filter(
-      ([key, value]) =>
-        ["id", "style", "class"].includes(key) ||
-        (storyData.argTypes[key]?.table?.category === "attributes" && !!value),
-    );
+    const attributes = Object.entries(elementArgs).filter(([key, value]) => {
+      const isBaseAttr = ["id", "style", "class"].includes(key);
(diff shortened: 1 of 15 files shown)


lubojr (approved):

santilland (approved):
LGTM 🌮

StefanBrand (commented):
I do not understand the implications enough to approve or reject this MR.

srijitcoder (approved):

baloola (approved):

PR #2327, opened 2026-05-12
feat: Export updateVectorLayerStyle function

A-Behairi:
Export the updateVectorLayerStyle function for external use.

- [x] All checks have passed

### DIFF
--- a/elements/layercontrol/src/main.js
+++ b/elements/layercontrol/src/main.js
@@ -38,7 +38,7 @@ addCommonStylesheet();
  *
  * ## Additional Helper Methods
  *
- * - `updateVectorLayerStyle`: Resolves a vector layer flat style object  by substituting any `["var", "<key>"]` placeholders with values from its `variables` property; imitating OL WebGL flat style variable substitution `webglLayer.updateStyleVariables`. 
+ * - `updateVectorLayerStyle`: Resolves a vector layer flat style object  by substituting any `["var", "<key>"]` placeholders with values from its `variables` property; imitating OL WebGL flat style variable substitution `webglLayer.updateStyleVariables`.
  *
  * Usage: `import { updateVectorLayerStyle } from "@eox/layercontrol";`
  *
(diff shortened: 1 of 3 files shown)


lubojr (approved):

silvester-pari (changes requested):
Thanks, where can we best document this? See the other usable methods documentation at https://github.com/EOX-A/EOxElements/blob/main/elements/map/src/main.js#L119-L134

A-Behairi:
I see that the documented exported helpers are from the root package, but in this case it will be exported as part of the source code, for it to be imported by external apps it would be like this:

```js
import { updateVectorLayerStyle } from "@eox/layercontrol/src/methods/layer-config/apply-updated-styles"
```
I would say its currently automatically documented through the types, but we can also include it as part of the layercontrol root exports if this is also interesting to others. what do you think?

silvester-pari:
> I see that the documented exported helpers are from the root package, but in this case it will be exported as part of the source code, for it to be imported by external apps it would be like this:
> 
> ```js
> import { updateVectorLayerStyle } from "@eox/layercontrol/src/methods/layer-config/apply-updated-styles"
> ```
> 
> I would say its currently automatically documented through the types, but we can also include it as part of the layercontrol root exports if this is also interesting to others. what do you think?

I'd like to have it documented in the root docs page, which doesn't necessarily mean it has to be exported in the main.js file I guess, but at least mentioned in the JSDoc for main.js. Does that make sense?

My reasoning for this is if it is documented nowhere, it can easily happen that we change/rename this method, don't realize that this is used somewhere, and then break applications that use this. If it is part of the JSDoc then it is a bit more "official", of course it would be even better to add a test that directly tests the existance of this method, or similar.

Currently, it is just an internal function and can change at any time and it's not considered a breaking change. If we expose it to the "public" then it needs to be handled somehow differently I guess :thinking:

A-Behairi:
Valid point, thanks for the emphasis, I added the documentation and a small test. Generally we export a lot of functions internally and since we ship the source code as is in the package, any exported thing can be imported by any app and the apps would then take the risk of it changing as you mentioned. After consideration, In this case I think it would make more sense to export it from main and make sure its stable.

silvester-pari (approved):
Thank you for the additions! Since this is something new, I'd consider it a `feat` instead of a `fix`.

PR #2328, opened 2026-05-13
fix: avoid duplicate vector source loads from `EOxSelectInteraction`

A-Behairi:
## Implemented changes

This PR fixes 2  issues that cause vector layers to fetch their source twice when select interactions are used.

1. The call building `Group` children, where it dropped the `createInteractions` flag, so children always defaulted to `true`. When `updateLayer` builds a new Group with `false` for diffing, its children still ran their interactions, which attaches extra style layers to the real map and triggering a duplicate fetch.

2. `EOxSelectInteraction` shares the target source then immediately calls `setMap`. If the source is mid fetch, the new layer's first `prepareFrame` calls `loadFeatures` again — OL's `loadedExtentsRtree_` only updates on success, so the in-flight call doesn't dedupe and a second XHR fires.

- [x] All checks have passed


silvester-pari (approved):
Thank you! Looks like a tricky hunt to find those issues. Looking at the code it seems ok to me, ideally we'd also have a small test showcasing the fix if that's not too much effort.

A-Behairi:
indeed it was, thanks for the approval, test added ✅

santilland (approved):
Thank you for the research in finding this, testing the pull request which integrates this fix everything seems to be working!
The resolving of multiple loading of vector sources is really great, especially for big/slow assets!

Issue #2292, opened 2026-04-27
Datetime filter UI for narrow widths in Itemfilter


A-Behairi:
### Problem
When the Itemfilter component is displayed with a low width, the current timestamp filter `type: range` `format: date` does not render well and becomes difficult to use.

#### Preview
<img width="1798" height="1043" alt="Image" src="https://github.com/user-attachments/assets/709ec2e4-8fdd-4291-a3da-0b9194e8b349" />

### Proposed Solution
- Evaluate supporting a new datepicker range format with either:
  - The native HTML input type="datetime-local" with probably 2 inputs for the range (start and end)
  - Alternatively, we can consider integrating the Timecontrol datepicker for assigning a date/time range with one input

Generally, options should be accessible and handle min/max configuration and range assignments.

silvester-pari:
I wanted to implement the timecontrol datepicker, but then realized that the vanilla calendar pro doesn't have time range selection, only a hack where it needs to be spawned three times, once only date, and two times only with time. Not yet 100% happy with the result, even if it works:

<img width="342" height="594" alt="Image" src="https://github.com/user-attachments/assets/807c83da-69fb-4e2f-8149-07cd2b36b847" />

PR #2338, opened 2026-05-19
feat: picker UI for date range filter

silvester-pari:
## Implemented changes

This PR introduces a date picker UI to date range filtering by integrating the `@eox/timecontrol` library. When a filter object is designated as a date, the component now dynamically renders a time picker interface in place of the standard range slider, allowing users to pick ranges across calendar days.

#### Component & UI Enhancements

* **Conditional Time Control Rendering:** Updated `range.js` to conditionally render `<eox-timecontrol>` when the filter format is identified as a date.
* **Property Additions:** Added `.suggestions` passing to the range component to populate available time control options.
* **Styling Adjustments:** Added minor CSS structural fixes (`.range-wrapper`) in `style.eox.js`.

#### Logic & Core Methods

* **Date Range Array Expansion:** Refactored `get-selected-dates.js` in the timecontrol picker to populate and return an array of *all consecutive days* within a selected range, rather than just returning the start and end bounds.
* **Filter Reset Integration:** Updated `resetRangeMethod` to check for the existence of `<eox-timecontrol>` and programmatically reset its dates using `dayjs.unix` timestamps when a user clears filters.
* **Event Handling:** Hooked up the `@select` event from the time control picker to dispatch standard Unix timestamp values back to the range input handler.

#### Helpers & Enums

* **New Helper:** Created `isDate()` to cleanly check if a filter object format is set to `"date"`.
* **Standardized Formats:** Centralized the `DATE_TIME_FORMAT` string (`YYYY-MM-DD`) into the enums configuration to maintain consistent formatting across components.

#### Dependencies

* Added `@eox/timecontrol: 2.4.0` to the project's dependencies.

#### Follow-up / further enhancements:
- Display time #2339
- Make string editable #2340

## Screenshots/Videos

image

image

https://github.com/user-attachments/assets/c1317c0e-80df-4672-ac0b-a077c7f28083

- [x] All checks have passed

### DIFF
--- a/elements/itemfilter/package.json
+++ b/elements/itemfilter/package.json
@@ -40,7 +40,6 @@
   },
   "dependencies": {
     "@eox/elements-utils": "^1.1.0",
-    "@eox/timecontrol": "2.4.0",
     "@eox/ui": "^1.1.0",
     "@floating-ui/dom": "^1.6.8",
     "@turf/boolean-intersects": "^7.3.5",
@@ -55,6 +54,7 @@
   },
   "peerDependencies": {
     "@eox/layout": "*",
-    "@eox/map": ">=2.0.0 || >=2.1.0-dev.1"
+    "@eox/map": ">=2.0.0 || >=2.1.0-dev.1",
+    "@eox/timecontrol": ">=2.0.0"
   }
 }
--- a/elements/itemfilter/package.json
+++ b/elements/itemfilter/package.json
@@ -40,6 +40,7 @@
   },
   "dependencies": {
     "@eox/elements-utils": "^1.1.0",
+    "@eox/timecontrol": "2.4.0",
     "@eox/ui": "^1.1.0",
     "@floating-ui/dom": "^1.6.8",
     "@turf/boolean-intersects": "^7.3.5",
--- a/elements/itemfilter/src/components/filters/range.js
+++ b/elements/itemfilter/src/components/filters/range.js
@@ -122,20 +122,16 @@ export class EOxItemFilterRange extends LitElement {
                 },
               ]}
               .initDate=${[
-                dayjs.unix(
-                  this.filterObject.state.min || this.filterObject.min,
-                ),
-                dayjs.unix(
-                  this.filterObject.state.max || this.filterObject.max,
-                ),
+                dayjs(this.filterObject.state.min || this.filterObject.min),
+                dayjs(this.filterObject.state.max || this.filterObject.max),
               ]}
               @select=${(e) =>
                 rangeInputHandlerMethod(
                   new CustomEvent("values", {
                     detail: {
                       values: [
-                        dayjs(e.detail.date[0]).unix(),
-                        dayjs(e.detail.date[1]).unix(),
+                        dayjs(e.detail.date[0]).valueOf(),
+                        dayjs(e.detail.date[1]).valueOf(),
                       ],
                     },
                   }),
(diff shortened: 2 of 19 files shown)


lubojr:
This is amazing addition which will really improve UX of the ItemFilter! Thank you very much for this!

Just wanted to confirm - the `suggestions` array will add the "dots" for the timecontrol day entries where there are any results?

lubojr (approved):

silvester-pari:
> This is amazing addition which will really improve UX of the ItemFilter! Thank you very much for this!
> 
> Just wanted to confirm - the `suggestions` array will add the "dots" for the timecontrol day entries where there are any results?

Now it does :innocent: 

image

I considered also adding the item popup, but I have to check if it is actually useful, might need to change the example story.

A-Behairi (commented):
This looks really great! thank you for this, I would suggest to close #2292

I tried it in eodash and there's one small issue, the filters in the external filters parameters for the date range returns the following:
```json
{
  "type": "range",
  "dirty": true,
  "key": "datetime",
  "format": "date",
  "title": "Date",
  "state": {
    "min": 1779660000,
    "max": 1779919200
  },
  "stringifiedState": "2026-05-25 - 2026-05-28"
}
```
the `stringifiedState` returns the correct state but `state.min` and `state.max` return the equivalent to `1970-01-21T14:21:00.000Z` and `1970-01-21T14:25:19.200Z`

for context the configuration object there is:
```json
{
  "key": "datetime",
  "title": "Date",
  "type": "range",
  "format": "date"
}
```

A-Behairi:
It seems that it's a Unix timestamp and not JS, so `new Date(result.state.min * 1000)` returns the expected value

silvester-pari:
> This looks really great! thank you for this, I would suggest to close #2292
> 
> I tried it in eodash and there's one small issue, the filters in the external filters parameters for the date range returns the following:
> 
> ```json
> {
>   "type": "range",
>   "dirty": true,
>   "key": "datetime",
>   "format": "date",
>   "title": "Date",
>   "state": {
>     "min": 1779660000,
>     "max": 1779919200
>   },
>   "stringifiedState": "2026-05-25 - 2026-05-28"
> }
> ```
> 
> the `stringifiedState` returns the correct state but `state.min` and `state.max` return the equivalent to `1970-01-21T14:21:00.000Z` and `1970-01-21T14:25:19.200Z`
> 
> for context the configuration object there is:
> 
> ```json
> {
>   "key": "datetime",
>   "title": "Date",
>   "type": "range",
>   "format": "date"
> }
> ```

Thank you @A-Behairi for testing it and spotting an issue! I fixed it in https://github.com/EOX-A/EOxElements/pull/2338/commits/b10da764853d99b15b39bfde68eb71a34bb29d6e, now the date should be returned in milliseconds as it was before.

A-Behairi (approved):

silvester-pari:
@lubojr I tried showing the dots also for time ranges (itemfilter allows range-filtering timestamps but also time ranges) and rendering a dot on each day, but this requires quite some further changes in the date picker and also does not 100% feel intuitive, since it's not clear that the repeating dots are supposed to be ranges :thinking: 

image

I think I will skip this for the first implementation and only include "dot" support for singe time stamps, and then we see if we require support for ranges as well. What do you think?

lubojr:
Yes, completely fine to skip this for ranges completely I think, right @A-Behairi? (we are aiming to fix the UX for EodashItemCatalog in the EOPF deployment by this, so it would be great to have the final OK based on that deployment).

A-Behairi:
Agreed, this is already a great addition

PR #2409, opened 2026-06-16
fix: timeline not updating on group layer changes

A-Behairi:
## Implemented changes

The timeline only reacted to top-level OL layer add/remove. Layers added or updated inside an existing group (e.g. eodash's mosaic in AnalysisGroup) didn't fire these events. This changes the listener to eox-map's `layerschanged` instead, with cleanup on disconnect. 

Note: re-init is now driven only by `.layers` changes, not direct OL mutation.

- [x] All checks have passed


silvester-pari (approved):
Thank you, had a quick look and looks good to me! That direct ol listener was probably still a remnant of a past setup where we directly connected to ol, but listening for eox-map updates is probably better in any case. Our `layers` array should anyway be the source of truth.

PR #2445, opened 2026-07-15
fix: layerConfig removeProperties behaviour

A-Behairi:
## Implemented changes

This reverts the layerConfig's `removeProperties` behaviour to drop the values of the tile URL initially only instead of on every update.

An example use case for this would be if the tile server exposes 2 params (`variables` and `expressions`) and the initial URL includes `variables` and the `layerConfig` updates `expressions` but the server gives `variables` precedence and do not render `expressions` unless `variables` is not present.
In that case the schema options can hold `removeProperties: ["variables"]` so that `expressions` would be passed on initial load of the form.

- [x] All checks have passed


lubojr (approved):
The usecase makes sense and my previous removal would not allow that behavior. Thank you for the fix!

PR #2447, opened 2026-07-17
fix: seed `layerConfig` start values for anyOf/oneOf/allOf schemas

lhoupert:
Opening the layer config panel can silently change what a layer shows. When the form schema keeps its fields inside `anyOf`/`oneOf`/`allOf` groups (the "pick one variant" style) instead of plain `properties`, `getStartVals` finds nothing, so the form falls back to the schema's defaults,  and those defaults get written back into the layer's tile URL. 

We hit this downstream (EOPF-Explorer/data-pipeline#348): opening the panel on an *ascending* Sentinel-1 layer flipped it to the form's *descending* default. This PR makes `getStartVals` also look inside those groups, so the form opens showing what the layer is actually rendering.

## For reviewers

I'm not a maintainer and not a JavaScript specialist, so please check the behaviour decisions rather than assume them:

- `anyOf`/`oneOf`/`allOf` forms now receive start values where they previously got none — a real change if anything relied on schema defaults winning; happy to gate it.
- A schema's own `properties` always beat same-named entries from the branches (regression-tested).
- A number field whose URL value isn't numeric (e.g. `rescale=auto`) now seeds the raw string instead of `NaN` — this also affects plain `properties` forms.
- `$ref`-based branches are still not handled (unchanged from before).

## Author attestation

- [x] I am a human, these are my changes, and I have reviewed and understood every change and can explain why each is correct.

AI-assisted (Claude): the investigation, fix and tests were AI-written, then checked by a second AI review pass. I verified the claims myself: each new test fails without its fix and passes with it, the full `layercontrol` suite is green (32 passing), and the ascending/descending flip above is real output from a live layer — not an illustration.

---

- [x] All checks have passed

### DIFF
--- a/elements/layercontrol/src/helpers/get-start-vals.js
+++ b/elements/layercontrol/src/helpers/get-start-vals.js
@@ -1,4 +1,29 @@
 /**
+ * Collects a schema's field definitions, from `properties` and from any `anyOf`/`oneOf`/`allOf`
+ * branches.
+ *
+ * json-editor uses those combinators for "pick one variant" forms, where the fields live in the
+ * branches and the schema's own `properties` is empty — so reading `properties` alone finds
+ * nothing.
+ *
+ * @param {{[key: string]: any}} schema - The schema object to collect fields from.
+ * @returns {{[key: string]: any}} - Field definitions keyed by property name.
+ */
+function collectFieldSchemas(schema) {
+  if (!schema || typeof schema !== "object") return {};
+
+  const fields = { ...schema.properties };
+  for (const key of ["anyOf", "oneOf", "allOf"]) {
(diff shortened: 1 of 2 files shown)


A-Behairi:
Thanks, @lhoupert, for the contribution. This is a nice enhancement that improves support for JSON schemas. If you’re interested in taking this further, I’d suggest extending`collectFieldSchemas` to resolve `$ref` against the schema's definitions/$defs, but I can also pick it up from here.

lhoupert:
Thank you! I would feel more confortable if you take it from here :-)

silvester-pari (approved):
Thank you both, great improvement!

PR #2472, opened 2026-08-04
Fix: jsonform round based on step for minmaxslider

lubojr:
## Implemented changes

Enhances the minmax custom editor in eox-jsonform by automatically calculating the decimal precision for the range slider based on the provided step attribute.

 Previously, the underlying slider used a default rounding of 2 decimal places, which caused higher-precision values (e.g., 0.005) to be rounded incorrectly.

### Changes:
- Added getDecimalPlaces helper to determine precision from step (supporting decimals and scientific notation).
- Replaced || with ?? when handling numeric properties to correctly support 0 as a valid value.
- Added a test

- [x] All checks have passed

### DIFF
--- a/elements/layercontrol/src/enums/stories/index.js
+++ b/elements/layercontrol/src/enums/stories/index.js
@@ -340,6 +340,7 @@ const LAYERCONFIG_LAYER_SEE = {
             maximum: 5,
             format: "range",
             default: 2,
+            step: 0.001,
           },
           vmax: {
             type: "number",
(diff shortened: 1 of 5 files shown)


elements/layercontrol/src/enums/stories/index.js
srijitcoder, line 351: Shouldn't we add one `step` in `vmax` too?

srijitcoder (approved):

elements/layercontrol/src/enums/stories/index.js
lubojr: Our code takes the step from either min or from max, so I don't think there is need for it.

PR #2474, opened 2026-08-04
fix: robust FlatGeoBuf bbox calculation for poles and antimeridian crossings

lubojr:
## Implemented changes

This PR improves the data fetching logic for the FlatGeoBuf source rect bbox for global, polar, and wrap-around map views.                                                                                                                                                         
                                                                                                                                                                
Previously, the bounding box for FlatGeoBuf requests was calculated (in our integration) by transforming only the corners of the map current view extent. This often failed to capture the true footprint in EPSG:4326 when crossing the poles or the antimeridian, resulting in data not being requested if the view contained e.g. north pole.
                                                                                                                           
### Key Changes:                                                                                                                                                   
- Densified Extent Transformation: Updated transformExtent helper to support "stops" (sampling points), allowing OpenLayers to accurately calculate the transformed bounding box for non-linear projections. (not sure if really needed but keeping it)
- Polar Crossing Detection: Implemented explicit checks for North/South Pole. If a pole is within the map extent, the request bounding box is expanded to cover the whole crs extent (of that pole)
- Antimeridian Wrap-around Support: Added logic to detect antimeridian crossings by comparing normalized longitudes of the extent edges. When a crossing is detected, the request expands to a full global longitude range (-180 to 180).                                                                                                                     
 - Added a unit test for both methods.

- [x] All checks have passed

### DIFF
--- a/elements/map/src/custom/sources/FlatGeoBuf.js
+++ b/elements/map/src/custom/sources/FlatGeoBuf.js
@@ -1,4 +1,5 @@
-import { transformExtent } from "ol/proj";
+import { transformExtent, transform } from "../../helpers/transform";
+import { containsCoordinate } from "ol/extent";
 import { deserialize } from "flatgeobuf/lib/mjs/geojson";
 import Vector from "ol/source/Vector.js";
 import GeoJSON from "ol/format/GeoJSON";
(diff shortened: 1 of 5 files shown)


elements/map/src/custom/sources/FlatGeoBuf.js
```diff
@@ -32,12 +33,68 @@ class FlatGeoBuf extends Vector {
...
+        transformedExtent[2] = 180;
+        if (containsNorthPole) {
+          transformedExtent[3] = 90;
+        }
+        if (containsSouthPole) {
+          transformedExtent[1] = -90;
+        }
+      }
+
+      // Antimeridian check
```
RobertOrthofer, line 81: I think the the same "point densification" as for the pole-check above would make sense here. Polar projections are confusing, but I think `[-1500, 500, -500, 1500]` in `EPSG:3413` should be an antimeridian-crossing extent that does not contain the pole and is not caught in the current antimeridian-check.

elements/map/test/cases/layer-type/test-flatgeobuf-polar-logic.js
```diff
@@ -0,0 +1,45 @@
...
+  expect(poleBbox.maxY).to.equal(90, "North pole crossing sets maxY to 90");
+
+  // 2. Test Antimeridian Crossing
+  // In EPSG:3857, the antimeridian is around 20037508
+  const antimeridianExtent = [20037000, 0, 20038000, 1000];
+  const webMercator = new Projection({ code: "EPSG:3857" });
+
+  const antiBbox = source.fgbBoundingBox(antimeridianExtent, webMercator);
+
+  expect(antiBbox.minX).to.equal(
```
RobertOrthofer: Maybe a "negative"-test would be good here, to make sure that a possible regression does not introduce false extent growings of a "normal" extent

elements/map/src/custom/sources/FlatGeoBuf.js
```diff
@@ -32,12 +33,68 @@ class FlatGeoBuf extends Vector {
...
+    if (sourceProjectionCode !== dataProjectionCode) {
+      // Check if the extent contains the poles in polar projections
+      // Transform pole coordinates to source projection
+      const poleNorth = transform([0, 90], "EPSG:4326", sourceProjectionCode);
+      const poleSouth = transform([0, -90], "EPSG:4326", sourceProjectionCode);
+
+      const containsNorthPole = containsCoordinate(extent, poleNorth);
+      const containsSouthPole = containsCoordinate(extent, poleSouth);
+
+      if (containsNorthPole || containsSouthPole) {
```
RobertOrthofer, line 68: as a super-edge-case, an extent that just "touches" a pole will also be expanded to the full longitude, but I doubt this will have any real-world relevancy. I think this if fine as is

RobertOrthofer (commented):
Some minor comments from my side, but all in all this looks good to me.

RobertOrthofer:
There is one more thing, which actually would result in a performance boost: 

When "expanding" the extent, we don't actually need to expand it over the entire world. Instead we can use multiple, smaller extents and call `deserialize` with each of those extents. 

For antimeridian crossing, his would give 2 extents: 

 - Extent A: [trueLeftLon, minY, 180, maxY]
 - Extent B: [-180, minY, trueRightLon, maxY]
 
For pole intersections, this would result in 4 extents, basically the 4 "wedges" up  (or down) to the pole, one for each quadrant. The latitude could be different for each quadrant, so that the extent gets as small as possible for each quadrant, if that makes sense...

There is a [getIntersection](https://openlayers.org/en/latest/apidoc/module-ol_extent.html#.getIntersection) function of `extent` by openlayers, so this could be done using relatively little logic here. 

While these changes (especially the antimeridian crossing) would be a cleaner implementation and result in a performance boost, the current solution is also working, so imo this *could* also be done in a separate PR.

lubojr:
Thank you very much for careful review! Really appreciated that you thought of the edge cases as well! I've integrated all comments from it except for the `pole intersections split` into 4 parts due (based on my guess) to higher a complexity of the code and would leave it as is for now. Tested in the integrating app that the logic over the north pole and antimeridian (where we now need it in the integration) still works.

elements/map/src/custom/sources/FlatGeoBuf.js
lubojr: I'd let this be as is.

RobertOrthofer (approved):
Looking good! And thanks for fixing the empty jsdoc-`returns`-documentation.

PR #2524, opened 2026-09-10
Fix/jsonform/configurable upload tooltip

lubojr:
##  Implemented changes                                                                                                                                            

Adds option to configure eox-jsonform ace editor toolbar upload custom tooltip text.

### Motivation

We need to give users extra text warning that their upload done via git-clerk GH integration will be visible only after a few seconds with current default text `![](source-gh-repo-link)`.                   
                                                                                                                                                                
 - [x] All checks have passed                                                                                                                                   
 - [x] The PR title follows conventional commit style and reflects the type of change (major/minor/patch, breaking)                                             
       (https://github.com/googleapis/release-please?tab=readme-ov-file#how-should-i-write-my-commits)                                                          

### DIFF
--- a/elements/jsonform/src/custom-inputs/ace-custom/createMarkdownToolbar.js
+++ b/elements/jsonform/src/custom-inputs/ace-custom/createMarkdownToolbar.js
@@ -256,7 +256,7 @@ export const createMarkdownToolbar = (editorInstance) => {
       ${
         hasUpload
           ? `
-      <button type="button" class="transparent no-round small" id="md-upload" title="Attach file">
+      <button type="button" class="transparent no-round small" id="md-upload" title="${uploadConfig.tooltip}">
         <i class="small">${icon(paperclipIcon)}</i>
         <span class="tooltip">${uploadConfig.tooltip}</span>
       </button>`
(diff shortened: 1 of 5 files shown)


silvester-pari (commented):
Thank you - maybe a dumb question but wouldn't this be better solved via a loading indicator? It would probably be good in any case to show some waiting/loading spinner while the file is uploaded, until it is inserted into the markdown. Would this generic solution be enough instead of the custom text?

lubojr:
It is not an ideal solution, but it is not when the upload is finished, but when github actually makes the raw link on the source repo available. At the time when link is inserted in markdown, this URL gives a 404, but it is the permanent one (alternative to the what we call a fallbackUrl which is instant, but gets obsolete/404 when the user would delete their fork repo permanently).

I don't think we can get this information from GH, that the repo raw image link is already available, as that is variable in the order of seconds - tens of seconds.

Essentialy the problem that we tried to solve with image section fallback url in storytelling renderer (but there just for whole ## sections containing image - e.g. image tour, hero image and hero video), but that one does not allow us to insert images that way into the map tour section text blocks for example.

lubojr:
Mitigating the root cause by switching the URL inplace after the file is reachable was the way to go instead. This PR is not needed anymore.

